PR #1, opened 2020-07-15
initial implementation

SaketramDurbha:
This is the initial, base implementation. A pared-down version of our prototype. Here's what it currently does:

- build container image and submit to Container Registry using reasonable defaults
- deploy to Cloud Run using reasonable default
- test GET / and expect 200 status code response
- delete Cloud Run service
- delete container image off Container Registry


cloudcontainerimage.go
```diff
@@ -0,0 +1,61 @@
+// Copyright 2020 Google LLC
```
kurtisvg: It's idiomatic for go to keep main / CLI entrypoints in a `/cmd` folder. I'd suggest structuring your project closer to some of these examples: 

https://github.com/golang-standards/project-layout/tree/master

Also, filenames are typically short, and usually use `snake_case`

cloudrunservice.go
```diff
@@ -0,0 +1,89 @@
...
+	"encoding/hex"
+	"fmt"
+	"log"
+)
+
+// Maximum length of a Cloud Run Service name.
+const maxCloudRunServiceNameLen = 53
+
+// CloudRunService represents a Cloud Run Service and holds its name and URL.
+type CloudRunService struct {
```
kurtisvg: Does this struct need to be accessed from outside the package? If not, keep it private.

cloudrunservice.go
```diff
@@ -0,0 +1,89 @@
...
+	// the associated Sample
+	sample *Sample
+
```
kurtisvg: nit: I would drop the spaces between these lines - they don't seem to be adding anything

cloudrunservice.go
```diff
@@ -0,0 +1,89 @@
...
+	"fmt"
+	"log"
+)
+
+// Maximum length of a Cloud Run Service name.
+const maxCloudRunServiceNameLen = 53
+
+// CloudRunService represents a Cloud Run Service and holds its name and URL.
+type CloudRunService struct {
+	// the associated Sample
```
kurtisvg: nit: This comment isn't really useful - it's pretty much restating what the code says (that whatever `Sample` is is associated with `CloudRunService`. Please remove or add new information.

cloudrunservice.go
```diff
@@ -0,0 +1,89 @@
...
+	sample *Sample
+
+	// the Service's name
+	name string
+
+	// the Service's root URL
+	url string
+}
+
+// newCloudRunService returns a pointer to a new CloudRunService with the provided name.
```
kurtisvg: nit: No need to clarify it's a pointer 
```suggestion
// newCloudRunService returns a new CloudRunService with the provided name.
```

cloudrunservice.go
```diff
@@ -0,0 +1,89 @@
...
+}
+
+// delete calls the external gcloud SDK and deletes the Cloud Run Service associated with the current CloudRunService.
+func (s *CloudRunService) delete() {
+	execCommand(gcloudCommandBuild([]string{
+		"run",
+		"services",
+		"delete",
+		s.name,
+		"--region=us-east4",
```
kurtisvg: Is region really supposed to be hard coded in?

cloudrunservice.go
```diff
@@ -0,0 +1,89 @@
...
+		"--platform=managed",
```
kurtisvg: Isn't this the default? I don't think we want to specify this here, because it will override any env vars that are set for other platforms.

cloudrunservice.go
kurtisvg: same for several other places

cloudrunservice.go
kurtisvg: same for several other places

cloudrunservice.go
```diff
@@ -0,0 +1,89 @@
...
+}
+
+// serviceName generates a Cloud Run service name for the provided sample. It concatenates the sample's name with a
+// random 10-character alphanumeric string.
+func serviceName(sample *Sample) string {
+	randBytes := make([]byte, cloudRunServiceNameRandSuffixLen/2)
+
+	_, err := rand.Read(randBytes)
+	if err != nil {
+		log.Panicf("Error generating crypto/rand bytes:  %v\n", err)
```
kurtisvg: In general, you never want to panic. Either handle the error or propagate it up to somewhere else to handle it.

config.go
```diff
@@ -0,0 +1,46 @@
...
+import (
+	"github.com/getkin/kin-openapi/openapi3"
+	"log"
+)
+
+const passResponseDescription = "PASS"
+
+// loadTestEndpoints loads a default test endpoint request (a GET / request expecting a 200 status code) into an
+// openapi3.Swagger object (see github.com/getkin/kin-openapi).
+func loadTestEndpoints() *openapi3.Swagger {
```
kurtisvg: should this be `loadTestEndpoints` or `defaultTestEndpoint`?

lifecycle.go
```diff
@@ -0,0 +1,159 @@
...
+// Phase represents one phase of a larger lifecycle. It has a list of goals that make up the actual step-by-step
+// execution of the phase in the form of exec.Cmd. In addition, each phase has a string ID that will uniquely identify
+// it in the context of its parent Lifecycle.
+type Phase struct {
+	id    string
+	goals []*exec.Cmd
+}
+
+// newLifeCycle builds a new Lifecycle given the provided phases and lifecycle id.
+func newLifeCycle(id string, phases []*Phase) *Lifecycle {
```
kurtisvg: Does this need to be a pointer?

lifecycle.go
```diff
@@ -0,0 +1,159 @@
...
+	lifecycle := Lifecycle{
+		id:     id,
+		phases: phases,
+	}
+
+	lifecycle.phaseMap = make(map[string]*Phase)
```
kurtisvg: Can you init this as part of the declaration above?

lifecycle.go
```diff
@@ -0,0 +1,159 @@
...
+	for _, phase := range phases {
```
kurtisvg: https://github.com/golang/go/wiki/CodeReviewComments#variable-names
```suggestion
	for _, p := range phases {
```

util.go
```diff
@@ -0,0 +1,64 @@
...
+)
+
+const cloudRunServiceNameRandSuffixLen = 10
+
+// gcloudCommandBuildcreates an exec.Cmd that calls the external gcloud executable. The exec.Cmd's name is set to
+// "gcloud" and the args are the ones provided in addition to a project flag and quiet flag.
+func gcloudCommandBuild(arg []string) *exec.Cmd {
+	var gcpProject string
+	if sample == nil {
+		gcpProject = os.Getenv("GOOGLE_CLOUD_PROJECT")
```
kurtisvg: Are we sure we want to set this, and not just relying on `gcloud`s project settings?

sample.go
```diff
@@ -0,0 +1,70 @@
...
+
+import (
+	"os"
+	"strings"
+)
+
+// Sample represents a Google Cloud Platform sample and associated properties.
+type Sample struct {
+	// The Google Cloud Project ID this sample will deploy to
+	googleCloudProject string
```
kurtisvg: nit:
```suggestion
	projectID string
```

sample.go
```diff
@@ -0,0 +1,70 @@
...
+
+	// The local directory this sample is located in
+	dir string
+
+	// The CloudRunService this sample will deploy to
+	cloudRunService *CloudRunService
```
kurtisvg: nit: (Alot of these names are somewhat long - Go tends to prefer shorter variables names)
```suggestion
	service *service
```

util.go
```diff
@@ -0,0 +1,64 @@
...
+	arg = append(arg, fmt.Sprintf("--project=%s", gcpProject), "--quiet")
+	cmd := exec.Command("gcloud", arg...)
+
+	return cmd
+}
+
+// execCommand executes an exec.Cmd. It redirects the commands stderr to this program's stderr and returns the output
+// in the form of a string.
+func execCommand(cmd *exec.Cmd) string {
+	if sample == nil {
```
kurtisvg: A comment here would be helpful - it's not clear to me what the difference between sampleDir and `sample.dir` are.

validateendpoints.go
```diff
@@ -0,0 +1,113 @@
...
+		success = success && validateEndpointOperation(pathItem.Put, endpoint, http.MethodPut)
+		success = success && validateEndpointOperation(pathItem.Trace, endpoint, http.MethodTrace)
+	}
+
+	return success
+}
+
+// validateEndpointOperation validates a single endpoint and a single HTTP method, and ensures that the request --
+// including the provided sample request body -- elicits the expected status code.
+func validateEndpointOperation(operation *openapi3.Operation, endpoint string, httpMethod string) bool {
```
kurtisvg: super nit: this seems like it would make more sense to be `endpoint string, operation *openapi3.Operation, httpMethod string` - usually you want to group related arguments together

validateendpoints.go
```diff
@@ -0,0 +1,113 @@
...
+		log.Printf("Testing %s endpoint\n", endpoint)
+		success = success && validateEndpointOperation(pathItem.Connect, endpoint, http.MethodConnect)
+		success = success && validateEndpointOperation(pathItem.Delete, endpoint, http.MethodDelete)
+		success = success && validateEndpointOperation(pathItem.Get, endpoint, http.MethodGet)
+		success = success && validateEndpointOperation(pathItem.Head, endpoint, http.MethodHead)
+		success = success && validateEndpointOperation(pathItem.Options, endpoint, http.MethodOptions)
+		success = success && validateEndpointOperation(pathItem.Patch, endpoint, http.MethodPatch)
+		success = success && validateEndpointOperation(pathItem.Post, endpoint, http.MethodPost)
+		success = success && validateEndpointOperation(pathItem.Put, endpoint, http.MethodPut)
+		success = success && validateEndpointOperation(pathItem.Trace, endpoint, http.MethodTrace)
```
kurtisvg: Can we use a for loop / slice here instead?

validateendpoints.go
```diff
@@ -0,0 +1,113 @@
...
+	return success
+}
+
+// validateEndpointOperation validates a single endpoint and a single HTTP method, and ensures that the request --
+// including the provided sample request body -- elicits the expected status code.
+func validateEndpointOperation(operation *openapi3.Operation, endpoint string, httpMethod string) bool {
+	if operation == nil {
+		return true
+	}
+	log.Printf("%s %s\n", httpMethod, endpoint)
```
kurtisvg: There are quite a few `logs` sprinkled in here. Are they debugging statements? There doesn't seem to be much context. I would suggest choosing which messages to expose to the end user with care, and use a different logger (or different library) for debugging (so it can be hidden until a user turns it on)

validateendpoints.go
```diff
@@ -0,0 +1,113 @@
...
+		log.Printf("%s: %s", mimeType, reqBodyStr)
+
+		reqBodyReader := strings.NewReader(reqBodyStr)
+		allTestsPassed = allTestsPassed && makeTestRequest(httpMethod, endpoint, mimeType, reqBodyReader, operation)
+	}
+
+	return allTestsPassed
+}
+
+// makeTestRequest makes a request based on the . It returns a success bool based on whether the returned status code
```
kurtisvg:
```suggestion
// makeTestRequest returns a success bool based on whether the returned status code
```

main.go
```diff
@@ -0,0 +1,86 @@
...
+		log.Fatalf("Error parsing sample direcotry: %v\n", err)
+	}
+
+	log.Println("Setting up configuration values")
+	sample = newSample(sampleDir)
+
+	log.Println("Loading test endpoints")
+	swagger := loadTestEndpoints()
+
+	log.Println("Activating service account")
```
kurtisvg: Are we sure we want to set this for every user? Why not just use whatever they've set for credentials

main.go
```diff
@@ -0,0 +1,86 @@
...
+	execCommand(gcloudCommandBuild([]string{
+		"auth",
+		"activate-service-account",
+		os.ExpandEnv("--key-file=${GOOGLE_APPLICATION_CREDENTIALS}"),
+	}))
+
+	log.Println("Building and deploying sample to Cloud Run")
```
kurtisvg: Should there be a flag to switch between Cloud Run / App Engine / Functions / etc?

main.go
```diff
@@ -0,0 +1,86 @@
...
+	sample.buildDeployLifecycle.execute()
```
kurtisvg: I'm not sure I understand why the Lifecycle object is needed to manage this process. It seems more complicated than necessary.

util.go
```diff
@@ -0,0 +1,64 @@
...
+	"os"
+	"os/exec"
+	"strings"
+)
+
+const cloudRunServiceNameRandSuffixLen = 10
+
+// gcloudCommandBuildcreates an exec.Cmd that calls the external gcloud executable. The exec.Cmd's name is set to
+// "gcloud" and the args are the ones provided in addition to a project flag and quiet flag.
+func gcloudCommandBuild(arg []string) *exec.Cmd {
```
kurtisvg: I think every instance of this function is wrapped with `execCommand` - does it really make sense for those to be two separate functions?

kurtisvg (changes requested):

cloudcontainerimage.go
tritone: +1 to the layout, and a 1-word filename is best if possible.

cloudcontainerimage.go
```diff
@@ -0,0 +1,61 @@
...
+package main
+
+import (
+	"fmt"
+	"os/exec"
+)
+
+// CloudContainerImage represents a sample's container image stored on the GCP Container Registry.
+type CloudContainerImage struct {
+	// the associated Sample
```
tritone: Comments should be full sentences with a period.

cloudcontainerimage.go
```diff
@@ -0,0 +1,61 @@
...
+		sample:       sample,
+		containerTag: cloudContainerImageTag(sample),
+	}
+
+	return &cloudContainerImage
+}
+
+// newCloudContainerImage creates a new CloudContainerImage for the provided Sample.
+func (c *CloudContainerImage) url() string {
+	return fmt.Sprintf("gcr.io/%s/%s", c.sample.googleCloudProject, c.containerTag)
```
tritone: Since c.sample is a pointer, you need nil-checking here. And/or you could have newCloudContainerImage error if a nil sample is passed in.

cloudcontainerimage.go
```diff
@@ -0,0 +1,61 @@
...
+
+package main
+
+import (
+	"fmt"
+	"os/exec"
+)
+
+// CloudContainerImage represents a sample's container image stored on the GCP Container Registry.
+type CloudContainerImage struct {
```
tritone: Given that you don't pass in any additional data to this type other than the sample itself, I'm wondering if this type is necessary. Could the url() method just be implemented on sample instead? Or are you planning on adding additional surface here?

cloudrunservice.go
tritone: +1, only export types and methods which need to be used from outside the package.

cloudrunservice.go
```diff
@@ -0,0 +1,89 @@
...
+	"crypto/rand"
+	"encoding/hex"
+	"fmt"
+	"log"
+)
+
+// Maximum length of a Cloud Run Service name.
+const maxCloudRunServiceNameLen = 53
+
+// CloudRunService represents a Cloud Run Service and holds its name and URL.
```
tritone: This could be more descriptive-- I don't think it corresponds to anything in the docs for Cloud Run so you need to give more context. Looking at this a bit it seems like this is intended for a single run of a single sample?

cloudrunservice.go
tritone: +1

util.go
```diff
@@ -0,0 +1,64 @@
...
+	}
+
+	cmd.Stderr = os.Stderr
+
+	log.Println("Executing ", cmd.String())
+
+	b, err := cmd.Output()
+	if err != nil {
+		fmt.Println(string(b))
+		log.Panicf("Error with exec cmd: %v\n", err)
```
tritone: Again, this should return an error to the caller rather than panicking. This can then be propagated up the call chain.

cloudcontainerimage.go
```diff
@@ -0,0 +1,61 @@
...
+	return &cloudContainerImage
+}
+
+// newCloudContainerImage creates a new CloudContainerImage for the provided Sample.
+func (c *CloudContainerImage) url() string {
+	return fmt.Sprintf("gcr.io/%s/%s", c.sample.googleCloudProject, c.containerTag)
+}
+
+// delete deletes the container image off of the Container Registry.
+func (c *CloudContainerImage) delete() {
```
tritone: I would expect this function (and others) to return an error in the case of failure.

config.go
```diff
@@ -0,0 +1,46 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package main
+
+import (
+	"github.com/getkin/kin-openapi/openapi3"
```
tritone: Have we checked on whether this dependency is okay to use?

lifecycle.go
```diff
@@ -0,0 +1,159 @@
...
+
+// Lifecycle is a representation of a large process that needs to be executed, like building and deploying a sample to
+// Cloud Run. Each lifecycle is made up of a list of different phases -- like build, deploy, and post-deploy -- and each
+// phase is in turn made up of small goals that act as the steps of each phase.
+type Lifecycle struct {
+	// id is a unique identifier for this Lifecycle. Not currently used.
+	id string
+
+	// phases contains an order list of phases associated with this lifecycle.
+	phases []*Phase
```
tritone: Why use pointers here instead of Phase directly? Same for phaseMap.

lifecycle.go
```diff
@@ -0,0 +1,159 @@
...
+	// phaseMap is a map that maps IDs to their associated phase in the Phases slice.
+	phaseMap map[string]*Phase
+}
+
+// Phase represents one phase of a larger lifecycle. It has a list of goals that make up the actual step-by-step
+// execution of the phase in the form of exec.Cmd. In addition, each phase has a string ID that will uniquely identify
+// it in the context of its parent Lifecycle.
+type Phase struct {
+	id    string
+	goals []*exec.Cmd
```
tritone: Again, likely no need to use a pointer here as far as I can tell.

lifecycle.go
```diff
@@ -0,0 +1,159 @@
...
+// phase is in turn made up of small goals that act as the steps of each phase.
+type Lifecycle struct {
+	// id is a unique identifier for this Lifecycle. Not currently used.
+	id string
+
+	// phases contains an order list of phases associated with this lifecycle.
+	phases []*Phase
+
+	// phaseMap is a map that maps IDs to their associated phase in the Phases slice.
+	phaseMap map[string]*Phase
```
tritone: What's the benefit of having a separate phaseMap field?

I don't think I fully grok what's going on in this file yet, but are there a finite number of types of phases that can exist? If so, I think a better design might be something like:

```
type Lifecycle struct {
    id string
    
   phases []Phase
}

type Phase interface {
    // Add methods that should be implemented by a phase if you need them-- could delegate an
    // execute method here for example, or a method that will log the phase name, or...
}

type Build struct {
    goals []exec.Cmd
    ...
}

// Implement methods required by Phase on each type .
```

The idea is that a lifecycle would have a slice of phases, and then you can codify the types of phases using types which implement the Phase interface. This is a really typical Go pattern. You can also make constructors for each kind of phase based on input data.

On the other hand-- is a lifecycle ALWAYS going to consist of exactly a build phase, a deploy phase, and a post deploy phase, in that order? Then it might be better to avoid making this needlessly flexible. In this case you could do something like:

```
type Lifecycle struct {
    id string
    build []exec.Cmd
    deploy []exec.Cmd
    postDeploy []exec.Cmd

}
```

... which makes it really explicit about what will happen when.

Actually, I don't even see any distinction between what happens in the different phases in the code below. If that will continue to be the case, it could be further simplified to
```
type Lifecycle struct {
    id string
    cmds []exec.Cmd

}
```

In any case, this is a super long comment, but happy to talk offline about how to implement this in a way that makes sense.

lifecycle.go
```diff
@@ -0,0 +1,159 @@
...
+}
+
+// buildDefaultLifecycle builds a build and deploy command lifecycle with reasonable defaults for a non-Java
+// project. It uses `gcloud builds submit` for building the samples container image and submitting it to the container
+// and `gcloud run deploy` for deploying it to Cloud Run.
+func buildDefaultLifecycle(sample *Sample) *Lifecycle {
+	lifecycle := buildLifecycle()
+
+	gcrURL := sample.cloudContainerImage.url()
+	lifecycle.phaseMap["build"].goals = []*exec.Cmd{
```
tritone: The fact that you are using "build" and "deploy" as flag values here is a bit of a smell-- see my comment above about changing the types in this file to avoid.

lifecycle.go
```diff
@@ -0,0 +1,159 @@
...
+		},
+		{
+			id:    "post-deploy",
+			goals: nil,
+		},
+	})
+}
+
+// execute executes the goals for each phase.
+func (l *Lifecycle) execute() {
```
tritone: So is there actually any distinction between what phase a command is assigned to? If not you should just remove the idea of phases entirely.

main.go
```diff
@@ -0,0 +1,86 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package main
+
+import (
+	"github.com/spf13/cobra"
```
tritone: What's the story with this dependency? Why is it helpful?

main.go
```diff
@@ -0,0 +1,86 @@
...
+func main() {
+	rootCmd := &cobra.Command{
+		Use:   "sst [sample-dir]",
+		Short: "An end-to-end tester for GCP samples",
+		Args:  cobra.ExactArgs(1),
+		Run:   root,
+	}
+
+	if err := rootCmd.Execute(); err != nil {
+		log.Panicf("Error with cobra rootCmd Execution: %v\n", err)
```
tritone: Don't panic here.

main.go
```diff
@@ -0,0 +1,86 @@
...
+	log.Println("Activating service account")
+	execCommand(gcloudCommandBuild([]string{
+		"auth",
+		"activate-service-account",
+		os.ExpandEnv("--key-file=${GOOGLE_APPLICATION_CREDENTIALS}"),
+	}))
+
+	log.Println("Building and deploying sample to Cloud Run")
+	sample.buildDeployLifecycle.execute()
+	defer sample.cloudRunService.delete()
```
tritone: Could these defer statements come earlier? Where are these resources being created?

sample.go
tritone: +1

sample.go
```diff
@@ -0,0 +1,70 @@
...
+}
+
+// newSample creates a new Sample object for the sample located in the provided local directory.
+func newSample(dir string) *Sample {
+	sample := Sample{
+		googleCloudProject: os.Getenv("GOOGLE_CLOUD_PROJECT"),
+		dir:                dir,
+	}
+
+	sample.cloudRunService = newCloudRunService(&sample)
```
tritone: This pattern, where cloudRunService (and the other objects) both are fields of Sample and have a pointer back to their Sample, is not something that I've seen typically. It feels like things are not encapsulated very well. Is there a way to pass just the data that we need to these objects, rather than the sample itself? It might take some more thinking about what functionality needs to live where.

validateendpoints.go
```diff
@@ -0,0 +1,113 @@
...
+// was included in the provided openapi3.Operation expected responses.
+func makeTestRequest(httpMethod, endpoint, mimeType string, reqBodyReader *strings.Reader, operation *openapi3.Operation) bool {
```
tritone: All of these should return errors rather than bools.

sample.go
```diff
@@ -0,0 +1,70 @@
...
+
+	// The CloudRunService this sample will deploy to
+	cloudRunService *CloudRunService
+
+	// The CloudContainerImage that represents the location of this sample's build container image in the GCP Container
+	// Registry
+	cloudContainerImage *CloudContainerImage
+
+	// The Lifecycle for building and deploying this sample to Cloud Run
+	buildDeployLifecycle *Lifecycle
```
tritone: Why use pointers for these?

validateendpoints.go
```diff
@@ -0,0 +1,113 @@
...
+
+	req.Header.Add("Authorization", "Bearer "+identityToken)
+	req.Header.Add("content-type", mimeType)
+
+	resp, err := (*client).Do(req)
+	if err != nil {
+		log.Panicf("Error executing http request: %v\n", err)
+	}
+
+	body, err := ioutil.ReadAll(resp.Body)
```
tritone: Unchecked error.

tritone (commented):
Good start! I have a whole bunch of comments. Happy to talk more particularly about the design aspects; let me know if you'd like to meet.

util.go
SaketramDurbha: Every instance of gcloudCommandBuild is not wrapped with execCommand. https://github.com/GoogleCloudPlatform/serverless-sample-tester/pull/1/files#diff-95256f97482a6d6739fe6ee6897b4114R63 is an example. And vice versa, every instance of execCommand doesn't wrap gcloudCommandBuild. https://github.com/GoogleCloudPlatform/serverless-sample-tester/pull/1/files#diff-95256f97482a6d6739fe6ee6897b4114R30 is an example. So for right now, I think it makes sense to keep these two functions separate.

main.go
SaketramDurbha: Yeah, that makes sense. The plan was to hold off on making a flag until we had functionality for those environments implemented.

config.go
danielylee01: We've previously discussed using openapi with the team, so we assumed we were okay on that front - is there something else we should check?

main.go
SaketramDurbha: This is helping us currently manage the command line argument that the program requires. Eventually, we're planning to have it manage more arguments and flags.

We're also planning to allow users to set environment variables and add values to a small configuration file in the future, all of which we're planning to let github.com/spf13/viper manage and reconcile. cobra will allow us to bind flags to viper's configuration values as well.

config.go
SaketramDurbha: Right now, it's just loading a default test endpoint, but once the custom endpoint functionality gets merged, we're planning for this function to load those endpoints in as well.

cloudrunservice.go
SaketramDurbha: If we allowed this, we would need to implement additional functionality for managing other GCP resources -- such as GKE clusters -- right?

cloudrunservice.go
SaketramDurbha: An alternative to this cloud be that we could require users to set the gcloud `run/region` property. Is that what you're suggesting?

cmd/sst/main.go
```diff
@@ -0,0 +1,28 @@
+// Copyright 2020 Google LLC
```
kurtisvg: I don't think a "sst" package is particularly necessary - just "cmd" is fine.

internal/cmd/sst/cmd.go
```diff
@@ -0,0 +1,88 @@
...
+	"fmt"
+	"github.com/GoogleCloudPlatform/serverless-sample-tester/internal/sample"
+	"github.com/GoogleCloudPlatform/serverless-sample-tester/internal/util"
+	"github.com/spf13/cobra"
+	"log"
+	"path/filepath"
+)
+
+// NewCommand creates the root command.
+func NewCommand() *cobra.Command {
```
kurtisvg: This is command line stuff, so I'd expect it to be under the `cmd` pacakge.

internal/gcloud/cloudrunservice.go
```diff
@@ -0,0 +1,89 @@
+// Copyright 2020 Google LLC
```
kurtisvg: nit: `cloud_run_service.go`

internal/gcloud/cloudrunservice.go
```diff
@@ -0,0 +1,89 @@
...
+	"github.com/GoogleCloudPlatform/serverless-sample-tester/internal/util"
+	"strings"
+	"unicode"
+)
+
+// Maximum length of a Cloud Run Service name.
+const maxCloudRunServiceNameLen = 53
+
+// Length of the random suffix at the end of the Cloud Run Service name.
+const cloudRunServiceNameRandSuffixLen = 10
```
kurtisvg: nit: you can combine these into a singe block
```suggestion
const {
   // Maximum length of a Cloud Run Service name.
    maxCloudRunServiceNameLen = 53
    // Length of the random suffix at the end of the Cloud Run Service name.
    cloudRunServiceNameRandSuffixLen = 10
}
```

Additionally, your comment and variable names are pretty much the same things - you can either use shorter var names or remove the comments (Go tends to favor short but meaningful variable names - `maxNameLen` is probably proficient enough given the file is "cloud run service").

internal/gcloud/cloudrunservice.go
```diff
@@ -0,0 +1,89 @@
...
+func (s CloudRunService) Delete() (err error) {
+	_, err = util.ExecCommand(util.GcloudCommandBuild(
+		"run",
+		"services",
+		"delete",
+		s.Name,
+		"--platform=managed",
+	))
+
+	return
```
kurtisvg: You should probably add some context to this error before returning it:
```
return errors.Wrap(err, "failed to delete CloudRunService")
```
```suggestion
return errors.Wrap(err, "failed to delete CloudRunService")
```

internal/lifecycle/lifecycle.go
```diff
@@ -0,0 +1,91 @@
...
+
+import (
+	"fmt"
+	"github.com/GoogleCloudPlatform/serverless-sample-tester/internal/util"
+	"os"
+	"os/exec"
+)
+
+// Lifecycle is a list of ordered exec.Cmd that should be run to execute a certain process.
+type Lifecycle []*exec.Cmd
```
kurtisvg, line 25: Is it necessary to create a new type for this? Can we just use `[]*exec.Cmd`?

internal/gcloud/cloudrunservice.go
```diff
@@ -0,0 +1,89 @@
...
+)
+
+// Maximum length of a Cloud Run Service name.
+const maxCloudRunServiceNameLen = 53
+
+// Length of the random suffix at the end of the Cloud Run Service name.
+const cloudRunServiceNameRandSuffixLen = 10
+
+// CloudRunService represents a Cloud Run service and stores its parameters.
+type CloudRunService struct {
```
kurtisvg: Should we have a common service interface that the other runtimes should implement?

internal/gcloud/cloudrunservice.go
```diff
@@ -0,0 +1,89 @@
...
+
+	url, err := util.ExecCommand(util.GcloudCommandBuild(
+		"run",
+		"--platform=managed",
+		"services",
+		"describe",
+		s.Name,
+		"--format=value(status.url)",
+	))
+	s.url = url
```
kurtisvg:
```suggestion
	s.url, err := util.ExecCommand(util.GcloudCommandBuild(
		"run",
		"--platform=managed",
		"services",
		"describe",
		s.Name,
		"--format=value(status.url)",
	))
```

internal/gcloud/cloudrunservice.go
```diff
@@ -0,0 +1,89 @@
...
+
+	return url, err
+}
+
+// ServiceName generates a Cloud Run service name for the provided sample. It concatenates the sample's name with a
+// random 10-character alphanumeric string.
```
kurtisvg: The 10-character limit doesn't look like it's hardcoded, so don't let users rely on it. 
```suggestion
// random alphanumeric string.
```

internal/sample/sample.go
```diff
@@ -0,0 +1,115 @@
...
+	"fmt"
+	"github.com/GoogleCloudPlatform/serverless-sample-tester/internal/gcloud"
+	"github.com/GoogleCloudPlatform/serverless-sample-tester/internal/lifecycle"
+	"github.com/GoogleCloudPlatform/serverless-sample-tester/internal/util"
+	"os/exec"
+	"strings"
+	"unicode"
+)
+
+//
```
kurtisvg: Update or remove

internal/sample/sample.go
```diff
@@ -0,0 +1,115 @@
...
+	Dir string
+
+	// The cloudRunService this sample will deploy to.
+	Service gcloud.CloudRunService
+
+	// The lifecycle for building and deploying this sample to Cloud Run.
+	BuildDeployLifecycle lifecycle.Lifecycle
+
+	// The URL location of this sample's build container image in the GCP Container Registry.
+	cloudContainerImageURL string
```
kurtisvg, line 43: _super_ nit: IMO, whitespace in code should be meaningful. It should be use to separate blocks of similar code or related vars in the same way we use indents separate paragraphs. Similarly to paragraphs, blocks that are too short or too long tend to hurt readability.

internal/sample/sample.go
```diff
@@ -0,0 +1,115 @@
...
+
+	buildDeployLifecycle := lifecycle.GetLifecycle(dir, service.Name, cloudContainerImageURL)
+
+	return &Sample{
+		Name:                   name,
+		Dir:                    dir,
+		Service:                service,
+		BuildDeployLifecycle:   buildDeployLifecycle,
+		cloudContainerImageURL: cloudContainerImageURL,
+	}, nil
```
kurtisvg: nit: it's probably good to separate the init and return values here for readability - otherwise the nil is a somewhat easy to miss.

internal/util/command.go
```diff
@@ -0,0 +1,58 @@
...
+import (
+	"fmt"
+	"log"
+	"os"
+	"os/exec"
+	"strings"
+)
+
+// The directory in which all of the exec.Cmds will be run.
+var commandsDir string
```
kurtisvg: does this need to be a global state?

internal/util/command.go
```diff
@@ -0,0 +1,58 @@
...
+
+// SetCommandsDir is the setter method for commandsDir.
+func SetCommandsDir(d string) {
+	commandsDir = d
+}
+
+// GcloudCommandBuild creates an exec.Cmd that calls the external gcloud executable. The exec.Cmd's name is set to
+// "gcloud" and the args are the ones provided in addition to a project flag and quiet flag.
+func GcloudCommandBuild(arg ...string) *exec.Cmd {
+	arg = append(arg, "--quiet")
```
kurtisvg: Something to think about: should this be appended to the front? Will any gcloud commands have flags that are passed in somewhere else?

internal/util/command.go
```diff
@@ -0,0 +1,58 @@
...
+	cmd := exec.Command("gcloud", arg...)
+
+	return cmd
+}
+
+// ExecCommand executes an exec.Cmd. It redirects the command's stderr to this program's stderr and returns the output
+// in the form of a string. The command will be run in commandsDir.
+func ExecCommand(cmd *exec.Cmd) (out string, err error) {
+	cmd.Dir = commandsDir
+	cmd.Stderr = os.Stderr
```
kurtisvg: What about stdout?

internal/util/validateendpoints.go
```diff
@@ -0,0 +1,137 @@
+// Copyright 2020 Google LLC
```
kurtisvg: filename: `endpoints.go` maybe?

kurtisvg (changes requested):

cmd/sst/main.go
tritone: +1; also should have some kind of comment indicating what it's for if you do need this separately.

internal/gcloud/cloudrunservice.go
tritone: +1, I like this idea if it's likely that there would be other runtimes.

internal/sample/sample.go
tritone: I'll disagree a little bit in this specific case-- I feel like this is a pretty typical layout when you have a struct with mostly exported fields and want to have a doc for each

internal/sample/sample.go
```diff
@@ -0,0 +1,115 @@
...
+	"os/exec"
+	"strings"
+	"unicode"
+)
+
+//
+const maxCloudContainerImageTagLen = 53
+
+// Sample represents a Google Cloud Platform sample and associated properties.
+type Sample struct {
```
tritone, line 30: Just wanted to say that I'm happy with how you simplified the surface here-- this looks much better than what you had before.

Also, moving these types to an internal directory also was a good move probably.

internal/cmd/sst/cmd.go
```diff
@@ -12,75 +12,77 @@
...
-func root(cmd *cobra.Command, args []string) {
+// root is responsible for the root command. It handles the application flow.
+func root(cmd *cobra.Command, args []string) error {
 	// Parse sample directory from command line argument
-	var err error
-	sampleDir, err = filepath.Abs(filepath.Dir(args[0]))
+	sampleDir, err := filepath.Abs(filepath.Dir(args[0]))
 	if err != nil {
-		log.Fatalf("Error parsing sample direcotry: %v\n", err)
+		return err
```
tritone: You can wrap these errors to indicate where they are being thrown from-- this helps with debugging. 

i.e. `return fmt.Errorf("filepath.Abs: %v", err)`

Same for errors below in this file.

internal/gcloud/cloudrunservice.go
tritone: +1 to all this.

internal/gcloud/cloudrunservice.go
tritone: This is similar to my comment above-- using `errors.Wrap` is a great solution as long as this code only has to run on Go 1.13+. I know you (@SaketramDurbha ) had asked me about error patterns via chat as well-- another good read on the subject if you're considering wrapping is https://blog.golang.org/go1.13-errors .

internal/gcloud/cloudrunservice.go
```diff
@@ -0,0 +1,89 @@
...
+const cloudRunServiceNameRandSuffixLen = 10
+
+// CloudRunService represents a Cloud Run service and stores its parameters.
+type CloudRunService struct {
+	Name string
+	url  string
+}
+
+// Delete calls the external gcloud SDK and deletes the Cloud Run Service associated with the current cloudRunService.
+func (s CloudRunService) Delete() (err error) {
```
tritone: Just a note, you don't _have_ to use the pattern of naming your return variable-- you can also just return a specific error (or a tuple as the case may be). This is fine but just wanted to make sure you're aware.

internal/lifecycle/lifecycle.go
tritone, line 25: Think I commented on this before but why use the indirection of a pointer here when you could just use a slice of exec.Cmd directly?

internal/lifecycle/lifecycle.go
```diff
@@ -0,0 +1,91 @@
...
+
+// Execute executes the commands of a lifecycle.
+func (l Lifecycle) Execute() error {
+	for _, c := range l {
+		_, err := util.ExecCommand(c)
```
tritone: If c is a pointer you may have to check for nil.

internal/lifecycle/lifecycle.go
tritone: Seems reasonable to use a named type so that the execute method can be on the type -- but either way is fine.

internal/lifecycle/lifecycle.go
```diff
@@ -0,0 +1,91 @@
+// Copyright 2020 Google LLC
```
tritone, line 1: Use git mv for moving files-- makes it much easier to review than if it looks like this file is entirely new.

internal/util/validateendpoints.go
```diff
@@ -24,90 +24,114 @@ import (
...
-var identityToken string
+// test holds the openapi3.Operation and HTTP method associated with a single
+type test struct {
+	operation  *openapi3.Operation
+	httpMethod string
+}
 
-// validateEndpoints tests all paths (represented by openapi3.Paths) with all HTTP methods and given response bodies
+// ValidateEndpoints tests all paths (represented by openapi3.Paths) with all HTTP methods and given response bodies
 // and make sure they respond with the expected status code. Returns a success bool based on whether all the tests
```
tritone: Returning a bool here seems a little odd. Is a simple bool enough information for the caller to diagnose what happened with their test?

internal/lifecycle/lifecycle.go
```diff
@@ -0,0 +1,91 @@
...
+			return err
+		}
+	}
+
+	return nil
+}
+
+// GetLifecycle returns a lifecycle built with reasonable defaults based on whether the sample is java-based
+// (has a pom.xml) that doesn't have a Dockerfile or isn't.
+func GetLifecycle(sampleDir, serviceName, gcrURL string) Lifecycle {
```
tritone: Call this NewLifecycle or something similar?

internal/sample/sample.go
```diff
@@ -0,0 +1,115 @@
...
+
+	// The lifecycle for building and deploying this sample to Cloud Run.
+	BuildDeployLifecycle lifecycle.Lifecycle
+
+	// The URL location of this sample's build container image in the GCP Container Registry.
+	cloudContainerImageURL string
+}
+
+// NewSample creates a new sample object for the sample located in the provided local directory.
+func NewSample(dir string) (*Sample, error) {
```
tritone, line 47: At some point I imagine you're going to have to have options of some kind that can be passed in here?

tritone (commented):
Really nice improvements here. A few more comments but I think I should pass it off to a reviewer who has more familiarity with your project after this.

internal/util/command.go
SaketramDurbha: We're currently purposefully not connecting the command's stdout to a specific Writer so that we can extract it later through `cmd.Output`.

This is partly because Go currently doesn't allow for you tor redirect a command's output 
and later get it's output bytes later at the same time  (https://github.com/golang/go/blob/master/src/os/exec/exec.go#L536).

Thinking about this again, one idea I have to get around this is to do something like this:

```
var stdout bytes.Buffer
cmd.Stdout = io.MultiWriter(os.Stdout, &stdout)
```

and then use `cmd.Run()` instead of `cmd.Output()`. But this is only if we want to print out the command's stdout every time regardless of whether there's an error or not.

internal/lifecycle/lifecycle.go
SaketramDurbha: Per an offline discussion with you (@tritone), we're planning on leaving this as is for now. Mainly because the Go SDK `exec.Command` function returns an `*exec.Cmd`pointer — instead of an `exec.Cmd` itself.

internal/util/command.go
SaketramDurbha: The other reason is because there will be times, I think, where we don't want to print the stdout of a command, like when we call `gcloud auth print-identity-token`. So maybe there could be an argument specifying whether we should print the stodout of a command or not.

internal/cmd/cmd.go
```diff
@@ -0,0 +1,82 @@
...
+		return fmt.Errorf("[cmd.Root] building and deploying sample to Cloud Run: %w", err)
+	}
+
+	log.Println("Getting identity token for gcloud auhtorized account")
+	var identToken string
+
+	a := append(util.GcloudCommonFlags, []string{
+		"auth",
+		"print-identity-token",
+	}...)
```
kurtisvg: The variadic operator and slice cancel each other out.  
```suggestion
	a := append(util.GcloudCommonFlags, "auth", "print-identity-token")
```

kurtisvg: you do this in a few other places as well

kurtisvg (approved):

internal/sample/sample.go
SaketramDurbha: Yeah, that's the plan.

internal/gcloud/cloudrunservice.go
SaketramDurbha: Per an offline discussion with Kurtis (@kurtisvg), we're going to plan to do this in the future, when we're about to implement functionality for other runtimes. Not in this PR.

PR #2, opened 2020-08-04
implement README parsing

SaketramDurbha:
This PR implements functionality to parse terminal commands for building and deploying a sample to the Cloud Run fully managed platform out of the sample's README.md. Code blocks that include commands that the serverless-sample-tester would like to run should be immediately preceded with a comment including the tag `sst-run-linuxmacos`. For example:

````text
[//]: # ({sst-run-linuxmacos})
```bash
gcloud builds submit --tag=gcr.io/${GOOGLE_CLOUD_PROJECT}/run-mysql
```
````

This PR does **not** include:
-  Revised algorithm to replace Cloud Run service names in parsed terminal commands.
-  Customizable Cloud Run region and revised hierarchy for customization of Cloud Run region.

These changes will be made in (a) separate PR(s).

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,5 +43,12 @@ For example:
 gcloud builds submit --tag=gcr.io/${GOOGLE_CLOUD_PROJECT}/run-mysql
 ```
 
+Do not set the Cloud Run region you'd like to deploy to through the `--region` flag in the `gcloud run` commands.
+Instead, as mentioned above, do so by setting the `run/region` gcloud property.
+
+In addition to setting a default Cloud Run region, make sure to deploy to the fully platform on Cloud Run. You can
+achieve this by setting the `run/platform` gcloud property to `managed` or passing in the `--platform=managed` flag
+to your `gcloud run` commands.
+
 If code tags aren't added to your README, the program will fall back to reasonable defaults to build and deploy your
 sample to Cloud Run based on whether your sample is java-based and doesn't have a Dockerfile or isn't.
--- a/README.md
+++ b/README.md
@@ -33,12 +33,12 @@ If you'd like, make sure to include the following comment code tag immediately p
 the program should build and deploy your sample:
 
 ```text
-[//]: # ({sst-run-linuxmacos})
+[//]: # ({sst-run-unix})
 ```
 
 For example:
 ````text
-[//]: # ({sst-run-linuxmacos})
+[//]: # ({sst-run-unix})
 ```
 gcloud builds submit --tag=gcr.io/${GOOGLE_CLOUD_PROJECT}/run-mysql
 ```
--- a/README.md
+++ b/README.md
@@ -47,9 +47,9 @@ gcloud builds submit --tag=gcr.io/${GOOGLE_CLOUD_PROJECT}/run-mysql
 Do not set the Cloud Run region you'd like to deploy to through the `--region` flag in the `gcloud run` commands.
 Instead, as mentioned above, do so by setting the `run/region` gcloud property.
 
-In addition to setting a default Cloud Run region, make sure to deploy to the fully platform on Cloud Run. You can
-achieve this by setting the `run/platform` gcloud property to `managed` or passing in the `--platform=managed` flag
+In addition to setting a default Cloud Run region, make sure to deploy to the fully managed platform on Cloud Run. You
+can achieve this by setting the `run/platform` gcloud property to `managed` or passing in the `--platform=managed` flag
 to your `gcloud run` commands.
 
-If code tags aren't added to your README, the program will fall back to reasonable defaults to build and deploy your
-sample to Cloud Run based on whether your sample is java-based and doesn't have a Dockerfile or isn't.
+If comment code tags aren't added to your README, the program will fall back to reasonable defaults to build and deploy
+your sample to Cloud Run based on whether your sample is java-based and doesn't have a Dockerfile or isn't.
--- a/README.md
+++ b/README.md
@@ -33,12 +33,12 @@ If you'd like, make sure to include the following comment code tag immediately p
 the program should build and deploy your sample:
 
 ```text
-[//]: # ({t-run-linuxmacos})
+[//]: # ({sst-run-linuxmacos})
 ```
 
 For example:
 ````text
-[//]: # ({t-run-build-linuxmacos})
+[//]: # ({sst-run-build-linuxmacos})
 ```bash
 gcloud builds submit --tag=gcr.io/${GOOGLE_CLOUD_PROJECT}/run-mysql
 ```
--- a/README.md
+++ b/README.md
@@ -29,8 +29,8 @@ Make sure to authorize the gcloud SDK and set a default project and Cloud Run re
 default Cloud Run region can be set by setting the `run/region` gcloud property.
 
 ### README parsing
-If you'd like, make sure to include the following comment code tag immediately preceding code blocks to customize how
-the program should build and deploy your sample:
+If you'd like, make sure to include the following comment code tag in your README to customize how the program should
+build and deploy your sample:
 
 ```text
 [//]: # ({sst-run-unix})
@@ -43,6 +43,18 @@ For example:
 gcloud builds submit --tag=gcr.io/${GOOGLE_CLOUD_PROJECT}/run-mysql
 ```
 ````
+
+When parsing the README for custom build and deploy commands, the serverless sample tester will include any commands
+inside a code fence that is immediately preceded by a line containing `{sst-run-unix}`. You can use Markdown syntax
+(e.g. `[//]: # ({sst-run-unix})`) to include this line without making it visible when rendered.
+
+The parsed commands will not be run through a shell, meaning that the program will not perform any expansions,
+pipelines, redirections or any other functions that shells are responsible for. This also means that popular shell
+builtin commands like `cd`, `export`, and `echo` will not be available or may not work as expected.  
+
+However, any environment variables referenced in the form of `$var` or `${var}` will expanded. In addition, bash-style
+multiline commands (i.e. non-quoted backslashes at the end of a line that indicate a line continuation) will also be 
+supported. 
 
 Do not set the Cloud Run region you'd like to deploy to through the `--region` flag in the `gcloud run` commands.
 Instead, as mentioned above, do so by setting the `run/region` gcloud property.
--- a/README.md
+++ b/README.md
@@ -25,4 +25,5 @@ go build -o sst cmd/main.go
 ./sst [sample-dir]
 ```
 
-Make sure to authorize the gcloud SDK and set a default project before running this program.
+Make sure to authorize the gcloud SDK and set a default project and Cloud Run region before running this program. A
+default Cloud Run region can be set by setting the `run/region` gcloud property.
--- a/README.md
+++ b/README.md
@@ -39,9 +39,10 @@ the program should build and deploy your sample:
 For example:
 ````text
 [//]: # ({sst-run-linuxmacos})
-```bash
+```
 gcloud builds submit --tag=gcr.io/${GOOGLE_CLOUD_PROJECT}/run-mysql
 ```
+````
 
 Do not set the Cloud Run region you'd like to deploy to through the `--region` flag in the `gcloud run` commands.
 Instead, as mentioned above, do so by setting the `run/region` gcloud property.
--- a/README.md
+++ b/README.md
@@ -38,7 +38,7 @@ the program should build and deploy your sample:
 
 For example:
 ````text
-[//]: # ({sst-run-build-linuxmacos})
+[//]: # ({sst-run-linuxmacos})
 ```bash
 gcloud builds submit --tag=gcr.io/${GOOGLE_CLOUD_PROJECT}/run-mysql
 ```
--- a/README.md
+++ b/README.md
@@ -27,3 +27,21 @@ go build -o sst cmd/main.go
 
 Make sure to authorize the gcloud SDK and set a default project and Cloud Run region before running this program. A
 default Cloud Run region can be set by setting the `run/region` gcloud property.
+
+### README parsing
+If you'd like, make sure to include the following comment code tag immediately preceding code blocks to customize how
+the program should build and deploy your sample:
+
+```text
+[//]: # ({t-run-linuxmacos})
+```
+
+For example:
+````text
+[//]: # ({t-run-build-linuxmacos})
+```bash
+gcloud builds submit --tag=gcr.io/${GOOGLE_CLOUD_PROJECT}/run-mysql
+```
+
+If code tags aren't added to your README, the program will fall back to reasonable defaults to build and deploy your
+sample to Cloud Run based on whether your sample is java-based and doesn't have a Dockerfile or isn't.
--- a/internal/lifecycle/lifecycle.go
+++ b/internal/lifecycle/lifecycle.go
@@ -60,7 +60,7 @@ func NewLifecycle(sampleDir, serviceName, gcrURL string) (Lifecycle, error) {
 			return nil, fmt.Errorf("[lifecycle.NewLifecycle] parsing README.md: %w", err)
 		}
 
-		fmt.Printf("No code blocks immediately preceded by %s found in README.md", codeTag)
+		fmt.Printf("No code blocks immediately preceded by %s found in README.md\n", codeTag)
 	} else {
 		fmt.Println("No README.md found")
 	}
--- a/internal/lifecycle/lifecycle.go
+++ b/internal/lifecycle/lifecycle.go
@@ -15,6 +15,7 @@
 package lifecycle
 
 import (
+	"errors"
 	"fmt"
 	"github.com/GoogleCloudPlatform/serverless-sample-tester/internal/util"
 	"log"
@@ -45,16 +46,21 @@ func (l Lifecycle) Execute(commandsDir string) error {
 // NewLifecycle tries to parse the different options provided for build and deploy command configuration. If none of
 // those options are set up, it falls back to reasonable defaults based on whether the sample is java-based
 // (has a pom.xml) that doesn't have a Dockerfile or isn't.
-func NewLifecycle(sampleDir, serviceName, gcrURL string) Lifecycle {
+func NewLifecycle(sampleDir, serviceName, gcrURL string) (Lifecycle, error) {
 	readmePath := filepath.Join(sampleDir, "README.md")
 
 	if _, err := os.Stat(readmePath); err == nil {
 		lifecycle, err := parseREADME(readmePath, serviceName, gcrURL)
 		if err == nil {
-			log.Println("Using build and deploy commands found in README")
-			return lifecycle
+			log.Println("Using build and deploy commands found in README.md")
+			return lifecycle, nil
 		}
-		fmt.Printf("Failed parsing README: %v\n", err)
+
+		if !errors.Is(err, errNoREADMECommandsFound) {
+			return nil, fmt.Errorf("[lifecycle.NewLifecycle] parsing README.md: %v\n", err)
+		}
+	} else {
+		fmt.Println("No README.md found")
 	}
 
 	pomPath := filepath.Join(sampleDir, "pom.xml")
@@ -68,11 +74,11 @@ func NewLifecycle(sampleDir, serviceName, gcrURL string) Lifecycle {
 
 	if pomE && !dockerfileE {
 		log.Println("Using default build and deploy commands for java samples without a Dockerfile")
-		return buildDefaultJavaLifecycle(serviceName, gcrURL)
+		return buildDefaultJavaLifecycle(serviceName, gcrURL), nil
 	}
 
 	log.Println("Using default build and deploy commands for non-java samples or java samples with a Dockerfile")
-	return buildDefaultLifecycle(serviceName, gcrURL)
+	return buildDefaultLifecycle(serviceName, gcrURL), nil
 }
 
 // buildDefaultLifecycle builds a build and deploy command lifecycle with reasonable defaults for a non-Java
--- a/internal/lifecycle/lifecycle.go
+++ b/internal/lifecycle/lifecycle.go
@@ -47,10 +47,7 @@ func (l Lifecycle) Execute(commandsDir string) error {
 func NewLifecycle(sampleDir, serviceName, gcrURL string) Lifecycle {
 	readmePath := fmt.Sprintf("%s/README.md", sampleDir)
 
-	_, err := os.Stat(readmePath)
-	readmeExists := err == nil
-
-	if readmeExists {
+	if _, err := os.Stat(readmePath); err == nil {
 		lifecycle, err := parseREADME(readmePath, serviceName, gcrURL)
 		if err == nil {
 			log.Println("Using build and deploy commands found in README")
@@ -62,7 +59,7 @@ func NewLifecycle(sampleDir, serviceName, gcrURL string) Lifecycle {
 	pomPath := fmt.Sprintf("%s/pom.xml", sampleDir)
 	dockerfilePath := fmt.Sprintf("%s/Dockerfile", sampleDir)
 
-	_, err = os.Stat(pomPath)
+	_, err := os.Stat(pomPath)
 	pomExists := err == nil
 
 	_, err = os.Stat(dockerfilePath)
--- a/internal/lifecycle/lifecycle.go
+++ b/internal/lifecycle/lifecycle.go
@@ -56,9 +56,11 @@ func NewLifecycle(sampleDir, serviceName, gcrURL string) (Lifecycle, error) {
 			return lifecycle, nil
 		}
 
-		if !errors.Is(err, errNoREADMECommandsFound) {
-			return nil, fmt.Errorf("[lifecycle.NewLifecycle] parsing README.md: %v\n", err)
+		if !errors.Is(err, errNoREADMECodeBlocksFound) {
+			return nil, fmt.Errorf("[lifecycle.NewLifecycle] parsing README.md: %w", err)
 		}
+
+		fmt.Printf("No code blocks immediately preceded by %s found in README.md", codeTag)
 	} else {
 		fmt.Println("No README.md found")
 	}
--- a/internal/lifecycle/lifecycle.go
+++ b/internal/lifecycle/lifecycle.go
@@ -41,7 +41,6 @@ func (l Lifecycle) Execute(commandsDir string) error {
 	return nil
 }
 
-
 // NewLifecycle tries to parse the different options provided for build and deploy command configuration. If none of
 // those options are set up, it falls back to reasonable defaults based on whether the sample is java-based
 // (has a pom.xml) that doesn't have a Dockerfile or isn't.
--- a/internal/lifecycle/lifecycle.go
+++ b/internal/lifecycle/lifecycle.go
@@ -17,6 +17,7 @@ package lifecycle
 import (
 	"fmt"
 	"github.com/GoogleCloudPlatform/serverless-sample-tester/internal/util"
+	"log"
 	"os"
 	"os/exec"
 )
@@ -40,22 +41,38 @@ func (l Lifecycle) Execute(commandsDir string) error {
 	return nil
 }
 
-// NewLifecycle returns a lifecycle built with reasonable defaults based on whether the sample is java-based
+
+// NewLifecycle tries to parse the different options provided for build and deploy command configuration. If none of
+// those options are set up, it falls back to reasonable defaults based on whether the sample is java-based
 // (has a pom.xml) that doesn't have a Dockerfile or isn't.
 func NewLifecycle(sampleDir, serviceName, gcrURL string) Lifecycle {
+	readmePath := fmt.Sprintf("%s/README.md", sampleDir)
+
+	_, err := os.Stat(readmePath)
+	readmeExists := err == nil
+
+	if readmeExists {
+		if lifecycle, err := parseREADME(readmePath, serviceName, gcrURL); err == nil {
+			log.Println("Using build and deploy commands found in README")
+			return lifecycle
+		}
+	}
+
 	pomPath := fmt.Sprintf("%s/pom.xml", sampleDir)
 	dockerfilePath := fmt.Sprintf("%s/Dockerfile", sampleDir)
 
-	_, err := os.Stat(pomPath)
+	_, err = os.Stat(pomPath)
 	pomExists := err == nil
 
 	_, err = os.Stat(dockerfilePath)
 	dockerfileExists := err == nil
 
 	if pomExists && !dockerfileExists {
+		log.Println("Using default build and deploy commands for java samples without a Dockerfile")
 		return buildDefaultJavaLifecycle(serviceName, gcrURL)
 	}
 
+	log.Println("Using default build and deploy commands for non-java samples or java samples with a Dockerfile")
 	return buildDefaultLifecycle(serviceName, gcrURL)
 }
 
--- a/internal/lifecycle/lifecycle.go
+++ b/internal/lifecycle/lifecycle.go
@@ -52,10 +52,12 @@ func NewLifecycle(sampleDir, serviceName, gcrURL string) Lifecycle {
 	readmeExists := err == nil
 
 	if readmeExists {
-		if lifecycle, err := parseREADME(readmePath, serviceName, gcrURL); err == nil {
+		lifecycle, err := parseREADME(readmePath, serviceName, gcrURL)
+		if err == nil {
 			log.Println("Using build and deploy commands found in README")
 			return lifecycle
 		}
+		fmt.Printf("Failed parsing README: %v\n", err)
 	}
 
 	pomPath := fmt.Sprintf("%s/pom.xml", sampleDir)
--- a/internal/lifecycle/lifecycle.go
+++ b/internal/lifecycle/lifecycle.go
@@ -20,6 +20,7 @@ import (
 	"log"
 	"os"
 	"os/exec"
+	"path/filepath"
 )
 
 // Lifecycle is a list of ordered exec.Cmd that should be run to execute a certain process.
@@ -45,7 +46,7 @@ func (l Lifecycle) Execute(commandsDir string) error {
 // those options are set up, it falls back to reasonable defaults based on whether the sample is java-based
 // (has a pom.xml) that doesn't have a Dockerfile or isn't.
 func NewLifecycle(sampleDir, serviceName, gcrURL string) Lifecycle {
-	readmePath := fmt.Sprintf("%s/README.md", sampleDir)
+	readmePath := filepath.Join(sampleDir, "README.md")
 
 	if _, err := os.Stat(readmePath); err == nil {
 		lifecycle, err := parseREADME(readmePath, serviceName, gcrURL)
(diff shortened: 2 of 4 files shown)

### CODE REVIEW

File: README.md [hunk 1]
```diff
@@ -25,4 +25,31 @@ go build -o sst cmd/main.go
...
+[//]: # ({sst-run-linuxmacos})
+```
+gcloud builds submit --tag=gcr.io/${GOOGLE_CLOUD_PROJECT}/run-mysql
+```
+````
+
+Do not set the Cloud Run region you'd like to deploy to through the `--region` flag in the `gcloud run` commands.
+Instead, as mentioned above, do so by setting the `run/region` gcloud property.
+
+In addition to setting a default Cloud Run region, make sure to deploy to the fully platform on Cloud Run. You can
```
jsimonweb: ' ... fully platform ' 

should be

' ... fully managed platform '

README.md
```diff
@@ -25,4 +25,31 @@ go build -o sst cmd/main.go
...
+achieve this by setting the `run/platform` gcloud property to `managed` or passing in the `--platform=managed` flag
+to your `gcloud run` commands.
+
+If code tags aren't added to your README, the program will fall back to reasonable defaults to build and deploy your
```
jsimonweb: "If code tags ..."

update to be

"If comment code tags ..."

jsimonweb (approved):

internal/lifecycle/readme.go
```diff
@@ -0,0 +1,127 @@
...
+	"github.com/GoogleCloudPlatform/serverless-sample-tester/internal/util"
+	"os"
+	"os/exec"
+	"regexp"
+	"strings"
+)
+
+// The tag that should appear immediately before code blocks in a README to indicate that the enclosed commands
+// are to be used by this program for building and deploying the sample.
+var codeTag = "sst-run-linuxmacos"
```
kurtisvg: Does this tag need to include `linuxmacos`? Shouldn't gcloud commands be the same for either Windows or Unix?

internal/lifecycle/readme.go
```diff
@@ -0,0 +1,127 @@
...
+	}
+	defer file.Close()
+
+	scanner := bufio.NewScanner(file)
+
+	var lifecycle Lifecycle
+
+	inCode := false
+	for scanner.Scan() {
+		line := strings.TrimSpace(scanner.Text())
```
kurtisvg: Do we need to Trim every line? That's a lot of work when we are only using the changed line some of the time

internal/lifecycle/readme.go
```diff
@@ -0,0 +1,127 @@
...
+
+		if strings.Contains(line, "```") && inCode {
+			inCode = false
```
kurtisvg: Do we care about when we are inCode unless we are already in a codeTag?

README.md
```diff
@@ -25,4 +25,31 @@ go build -o sst cmd/main.go
...
-Make sure to authorize the gcloud SDK and set a default project before running this program.
+Make sure to authorize the gcloud SDK and set a default project and Cloud Run region before running this program. A
+default Cloud Run region can be set by setting the `run/region` gcloud property.
+
+### README parsing
+If you'd like, make sure to include the following comment code tag immediately preceding code blocks to customize how
+the program should build and deploy your sample:
+
+```text
+[//]: # ({sst-run-linuxmacos})
```
kurtisvg: What's the significance of `[//] #`?

internal/lifecycle/readme.go
```diff
@@ -0,0 +1,127 @@
...
+		} else if strings.Contains(line, codeTag) {
```
kurtisvg: Is `Contains` all we want? Shouldn't the line be pretty much exactly `[//] # SST_RUN_xxx` once trimmed?

internal/lifecycle/readme.go
```diff
@@ -0,0 +1,127 @@
...
+			scanner.Scan()
+			startCodeBlockLine := scanner.Text()
+
+			if !strings.Contains(startCodeBlockLine, "```") {
+				return nil, fmt.Errorf("[lifecycle.parseREADME] parsing README: incorrect format")
```
kurtisvg: Errors should be informative and ideally actionable. This say "incorrect format" - how does the user act to fix that?

internal/lifecycle/readme.go
```diff
@@ -0,0 +1,127 @@
...
+
+	scanner := bufio.NewScanner(file)
+
+	var lifecycle Lifecycle
+
+	inCode := false
+	for scanner.Scan() {
+		line := strings.TrimSpace(scanner.Text())
+
+		if strings.Contains(line, "```") && inCode {
```
kurtisvg: You should check `inCode` first - `Contains` is a more expensive operation

internal/lifecycle/readme.go
```diff
@@ -0,0 +1,127 @@
...
+			scanner.Scan()
+			startCodeBlockLine := scanner.Text()
+
+			if !strings.Contains(startCodeBlockLine, "```") {
+				return nil, fmt.Errorf("[lifecycle.parseREADME] parsing README: incorrect format")
+			}
+
+			inCode = true
+		} else if inCode {
+			for line[len(line)-1] == '\\' {
```
kurtisvg: A comment would be good here - I'm not sure I understand what's happening here

internal/lifecycle/readme.go
```diff
@@ -0,0 +1,127 @@
...
+	}
+
+	return lifecycle, nil
+}
+
+// replaceServiceName takes a terminal command string as input and replaces the Cloud Run service name, if any.
+// It detects whether the command is a gcloud run command and replaces the last argument that isn't a flag
+// with the input service name.
+func replaceServiceName(command, serviceName string) string {
+	containsGcloud, _ := regexp.MatchString(`\bgcloud\b`, command)
```
kurtisvg: Probably worth a comment on why we ignore these errors

internal/lifecycle/readme.go
```diff
@@ -0,0 +1,127 @@
...
+	containsRun, _ := regexp.MatchString(`\brun\b`, command)
+
+	if !(containsGcloud && containsRun) {
+		return command
+	}
+
+	sp := strings.Split(command, " ")
+	for i := len(sp) - 1; i >= 0; i-- {
+		if !strings.Contains(sp[i], "--") {
+			sp[i] = serviceName
```
kurtisvg, line 202: Can gcloud flags have args or are they always in the `flag=val` format?

```
gcloud run --flag1 arg1 serviceName --flag2 arg2
```

internal/lifecycle/readme.go
```diff
@@ -0,0 +1,127 @@
...
+	var lifecycle Lifecycle
+
+	inCode := false
+	for scanner.Scan() {
+		line := strings.TrimSpace(scanner.Text())
+
+		if strings.Contains(line, "```") && inCode {
+			inCode = false
+		} else if strings.Contains(line, codeTag) {
+			scanner.Scan()
```
kurtisvg: What if this is false?

internal/lifecycle/readme.go
```diff
@@ -0,0 +1,127 @@
...
+		return nil, err
+	}
+	defer file.Close()
+
+	scanner := bufio.NewScanner(file)
+
+	var lifecycle Lifecycle
+
+	inCode := false
+	for scanner.Scan() {
```
kurtisvg, line 140: You need to check `scanner.Err` I think?

internal/lifecycle/readme.go
kurtisvg: Is this to support multi line code?

internal/lifecycle/readme.go
```diff
@@ -0,0 +1,127 @@
...
+
+	if err := scanner.Err(); err != nil {
+		return nil, fmt.Errorf("[lifecycle.parseREADME] README bufio.Scanner: %w", err)
+	}
+
+	if len(lifecycle) == 0 {
+		return nil, fmt.Errorf("[lifecycle.parseREADME] parsing README: no commands found")
+	}
+
+	return lifecycle, nil
```
kurtisvg: nit: This whole function might be more readable if you allow for some nesting. Some example psuedo:
```
for line in scanner {
     // check scanner.Err
     line := scanner.Text()
     if line contains code_tag {
          for command in code_block {
                // parse command
                // convert command
           }
     }
}
```

kurtisvg (changes requested):

README.md
SaketramDurbha: This, in my understanding, is one syntax for a line comment in Markdown. It's the same syntax used here: https://raw.githubusercontent.com/googleapis/java-bigtable/master/README.md , for a similar purpose. 

However, formal documentation for Markdown comments has been hard to find. But it appears there are multiple syntaxes for line comments.

internal/lifecycle/readme.go
SaketramDurbha: As mentioned above (https://github.com/GoogleCloudPlatform/serverless-sample-tester/pull/2#discussion_r465461157), it appears there are multiple syntaxes for line comments in Mardown. That's why the liberal option of just looking for a mention the specific code tag in the line is used. I'm not sure whether we should just require this line specifically.

And if we stick with the initial implementation, after some thought, I'm now thinking that we shouldn't thrown an error if we aren't able to find the start of a code block on the next line; we should just ignore the current line with a code tag.

SaketramDurbha: Also, this, I believe, is the same way that the releasetool looks for package version update tags/markers in READMEs like the ones found here:
https://raw.githubusercontent.com/googleapis/java-bigtable/master/README.md

releasetool source: https://github.com/googleapis/releasetool/blob/master/releasetool/commands/start/java.py#L229

internal/lifecycle/readme.go
SaketramDurbha: I think it would be the same for Unix. But I wasn't not sure how the `exec.Command` and associated go interfaces.

One thing that concerns me specifically is that os.ExpandEnv seems to only expand variables with the `$VAR` or `${VAR}` syntax. Whereas it appears that Windows Command Prompt users can expand variables using the `%VAR%` syntax.

README.md
dmahugh: This syntax is widely supported and probably the right choice here, but comment syntax in Markdown isn't standardized, unfortunately. FYI, this SO thread has a lot of info: https://stackoverflow.com/questions/4823468/comments-in-markdown

internal/lifecycle/readme.go
dmahugh: Might want to just focus on bash syntax for now, since Windows syntax has a couple of variations. Windows CMD prompt expands environment variables with ```%varname%``` syntax, and PowerShell (which is more commonly used these days) expands them with ```$env:varname``` syntax.

dmahugh (approved):
LGTM (modulo comments from others), added some info on two topics.

internal/lifecycle/readme.go
SaketramDurbha: Will add a comment. Also an explanation here:

The only scenario in which the `regexp.MatchString` function will return an error is if the regex fails to compile. But in our case, since we're using the same two pieces of regex each time and we know that they're valid, it seemed like we could safely ignore these errors.

`regexp.MatchString` source: https://github.com/golang/go/blob/master/src/regexp/regexp.go#L548 .

Given this, I don't know if it's acceptable to be ignoring the errors or if we should be catching the errors for good measure anyways.

internal/lifecycle/readme.go
SaketramDurbha: I think the `bufio.Scanner` `Scan` method returns a bool indicating whether there was an error or at the end of file. If the bool indicates there is potentially an error, then we can check `Scanner.Err` to see what that error message specifically is.

Documentation here: https://golang.org/pkg/bufio/#Scanner.Scan .

README.md
kurtisvg: SG, but probably worth mentioning that in some fashion in the README

kurtisvg: To clarify, we should clarify what the user needs to do - is it a commented line? which kind of comment do we support? Can it be added somewhere else? We want to minimize the risk of a false positive with this.

internal/lifecycle/readme.go
kurtisvg: Looks like the source you've linked matches `{<some_tag>}` (not just `<some_tag>`). I think it would make sense to be more explicit (again to minimize false positives)

internal/lifecycle/readme.go
kurtisvg: A better alternative here would be to save the regex as a constant, so it's calculated at compile time. You can use `regex.MustCompile()` and then `(*Regexp).MatchString`. This will help you avoid the case where someone modifies the regex and you don't find out it's invalid until your compiler runs.

internal/lifecycle/readme.go
SaketramDurbha: Sorry, my mistake. I've update the `codeTag` constant to be `{sst-run-unix}`

internal/lifecycle/readme.go
SaketramDurbha: I believe they can have args and can be in the format followed in your example.

We (@danielylee01) were to revise the Cloud Run service name detection and replacing algorithm in a future PR so as to separate the work each of us do into separate PRs.

SaketramDurbha:
Per a recommendation from @kurtisvg , we've updated the code tag from `sst-run-linuxmacos` to `sst-run-unix`.

internal/lifecycle/readme.go
kurtisvg: SGTM - let's open an issue then for this PR to make sure we don't lose track. We can close it with a future PR once complete.

internal/lifecycle/readme.go
SaketramDurbha: Sorry, do you mean something like this?

```
r1 := regexp.MustCompile(`\bgcloud\b`)
r2 := regexp.MustCompile(`\brun\b`)

containsGcloud:= r1.MatchString(command)
containsRun := r2.MatchString(command)
```

kurtisvg: Yup!

README.md
SaketramDurbha: Okay, should we require code tags follow this regular expression pattern?:

```
\[//\]:\s+#\s+\(\s*{sst-run-linuxmacos}\s*\)
```

This would require a comment in the form of the `[//] # (comment)` syntax. We could also require that a code tag line following this format _immediately_ precede a code block.

internal/lifecycle/readme.go
SaketramDurbha: Okay, planning to open an issue entitled "Cloud Run service name detection and replacing algorithm flawed" after this current PR gets merged.

README.md
kurtisvg: To clarify, I think given the ambiguity of the comment syntax I agree with your initial decision to use a looser requirement. However, you should make sure that what you are looking for is clearly defined in the README. So something like:

> When parsing the README for custom build/deploy instructions, SST will include any commands inside a code fence that is immediately preceded by a line containing `{sst-run-unix}`. You can use a markdown comment syntax (e.g. `[//] # ({sst-run-unix})`) to include this line without making it visible when rendered.

This makes it clear that:
1. The line only has to contain `{sst-run-unix}`
2. There can't be any empty lines between the line containing the tag and the code block.
3. They can use `[//] #`, but the exact format is up to them. 

You may also which to clarify the following points:
* Are multiline commands supported?
* Is bash syntax supported (`cd` or `export VAR=VAL`)?

internal/lifecycle/lifecycle.go
```diff
@@ -40,22 +41,39 @@ func (l Lifecycle) Execute(commandsDir string) error {
...
+// NewLifecycle tries to parse the different options provided for build and deploy command configuration. If none of
+// those options are set up, it falls back to reasonable defaults based on whether the sample is java-based
 // (has a pom.xml) that doesn't have a Dockerfile or isn't.
 func NewLifecycle(sampleDir, serviceName, gcrURL string) Lifecycle {
+	readmePath := fmt.Sprintf("%s/README.md", sampleDir)
+
+	_, err := os.Stat(readmePath)
+	readmeExists := err == nil
+
+	if readmeExists {
```
kurtisvg: nit: I think you can remove this variable from scope altogether now 
```suggestion
	

	if _, err := os.Stat(readmePath); err != nil {
```

internal/lifecycle/readme.go
```diff
@@ -0,0 +1,149 @@
...
+				if err := scanner.Err(); err != nil && !s {
+					return nil, fmt.Errorf("[lifecycle.parseREADME] README bufio.Scanner: %w", err)
+				}
+
+				if !c {
+					return nil, fmt.Errorf("[lifecycle.parseREADME] parsing README: expecting start of code block immediately after code tag")
+				} else { // scanner.Scan falied, but no error found. EOF.
+					return nil, fmt.Errorf("[lifecycle.parseREADME] parsing README: unexpected EOF; file ended immediately after code tag")
+				}
+			}
```
kurtisvg: Can this be simplified? It looks like :
```
if !s or !c {
    if err and !s -> scanner err
    if c -> not code block err
    else -> EoF err
}
```

I think this is simplified as: 

```
if !s {
   if err -> scanner err
   else -> EoF err
}
if !c {
   -> not code block err
}
```

internal/lifecycle/readme.go
```diff
@@ -0,0 +1,149 @@
...
+				}
+
+				line = strings.TrimSpace(line)
+
+				// If there is a backslash at the end of the line, this is a multiline command. Keep scanning to get
+				// entire command.
+				for line[len(line)-1] == '\\' {
+					line = line[:len(line)-1]
+
+					scanner.Scan()
```
kurtisvg: What happens if this Scan is false? (or contains "```"?)

internal/lifecycle/readme.go
```diff
@@ -0,0 +1,149 @@
...
+
+	return lifecycle, nil
+}
+
+// replaceServiceName takes a terminal command string as input and replaces the Cloud Run service name, if any.
+// It detects whether the command is a gcloud run command and replaces the last argument that isn't a flag
+// with the input service name.
+func replaceServiceName(command, serviceName string) string {
+	r1 := regexp.MustCompile(`\bgcloud\b`)
+	r2 := regexp.MustCompile(`\brun\b`)
```
kurtisvg: nit: You should move these to global vars (so they are constants, and not re initialized each time).

also: better variables names if global constants

internal/lifecycle/readme.go
```diff
@@ -0,0 +1,149 @@
...
+
+	containsGcloud := r1.MatchString(command)
+	containsRun := r2.MatchString(command)
+
+	if !(containsGcloud && containsRun) {
```
kurtisvg: nit: You can avoid adding these variables to scope as well
```suggestion
	if !(containsGcloud.MatchString(command) && containsRun.MatchString(command)) {
```

kurtisvg (approved):

internal/lifecycle/lifecycle.go
```diff
@@ -40,9 +41,21 @@ func (l Lifecycle) Execute(commandsDir string) error {
 	return nil
 }
 
-// NewLifecycle returns a lifecycle built with reasonable defaults based on whether the sample is java-based
+// NewLifecycle tries to parse the different options provided for build and deploy command configuration. If none of
+// those options are set up, it falls back to reasonable defaults based on whether the sample is java-based
 // (has a pom.xml) that doesn't have a Dockerfile or isn't.
 func NewLifecycle(sampleDir, serviceName, gcrURL string) Lifecycle {
+	readmePath := fmt.Sprintf("%s/README.md", sampleDir)
```
tritone: You can use path/filepath for path manipulations (here and below). See https://gobyexample.com/file-paths and https://golang.org/pkg/path/filepath/

internal/lifecycle/lifecycle.go
```diff
@@ -40,9 +41,21 @@ func (l Lifecycle) Execute(commandsDir string) error {
...
+
+	if _, err := os.Stat(readmePath); err == nil {
+		lifecycle, err := parseREADME(readmePath, serviceName, gcrURL)
+		if err == nil {
+			log.Println("Using build and deploy commands found in README")
+			return lifecycle
+		}
+		fmt.Printf("Failed parsing README: %v\n", err)
```
tritone: I would think it would be preferable to have a fatal error here if the README exists but is invalid, rather than continuing to the backup methods.

internal/lifecycle/lifecycle.go
```diff
@@ -53,9 +66,11 @@ func NewLifecycle(sampleDir, serviceName, gcrURL string) Lifecycle {
 	dockerfileExists := err == nil
```
tritone: Technically there can be other errors from os.Stat other than the file not existing. It's probably fine not to distinguish but the variable names are slightly misleading.

internal/lifecycle/readme.go
```diff
@@ -0,0 +1,156 @@
...
+
+	gcloudCommandRegexp   = regexp.MustCompile(`\bgcloud\b`)
+	cloudRunCommandRegexp = regexp.MustCompile(`\brun\b`)
+
+	gcrURLRegexp = regexp.MustCompile(`gcr.io/.+/\S+`)
+)
+
+// parseREADME parses a README file with the given name. It reads terminal commands surrounded by one of the codeTags
+// listed above and loads them into a Lifecycle. In the process, it replaces the Cloud Run service name and Container
+// Registry tag with the provided inputs.
```
tritone, line 103: General comment looking at this function: parsing logic is pretty complicated and can be very tricky to get right in all the edge cases. How well-defined is the README format? Is doing this parsing a hard requirement for your project? If possible it would simplify things a lot if you could force users to provide input data as a json and then rely on the nice built in support that golang has for parsing those.

internal/lifecycle/readme.go
```diff
@@ -0,0 +1,156 @@
...
+			if s := scanner.Scan(); !s {
+				if err := scanner.Err(); err != nil && !s {
+					return nil, fmt.Errorf("[lifecycle.parseREADME] README bufio.Scanner: %w", err)
+				}
+				return nil, fmt.Errorf("[lifecycle.parseREADME] parsing README: unexpected EOF; file ended" +
+					"immediately after code tag")
+			}
+
+			startCodeBlockLine := scanner.Text()
+			c := strings.Contains(startCodeBlockLine, "```")
```
tritone: I'd also use a constant for the triple backticks (and any other magic values like this).

internal/lifecycle/readme.go
```diff
@@ -0,0 +1,156 @@
...
+	"os"
+	"os/exec"
+	"regexp"
+	"strings"
+)
+
+var (
+	// The tag that should appear immediately before code blocks in a README to indicate that the enclosed commands
+	// are to be used by this program for building and deploying the sample.
+	codeTag = "{sst-run-unix}"
```
tritone, line 30: Seems like these should be const.

internal/lifecycle/readme.go
```diff
@@ -0,0 +1,156 @@
...
+					if strings.Contains(l, "```") {
+						return nil, fmt.Errorf("[lifecycle.parseREADME] parsing README: unexpected end of" +
+							"code block; expecting command line continuation")
+					}
+
+					line = line + strings.TrimSpace(l)
+				}
+
+				line = os.ExpandEnv(line)
+				line = gcrURLRegexp.ReplaceAllString(line, gcrURL)
```
tritone: Is it possible to separate out the logic involved in just reading the readme from this text substitution? It would involve an additional pass over the data but I think it would make the code clearer (and it doesn't seem like efficiency is the biggest concern here). In this case you could do something like:

* Scan the file and generate a slice of strings that correspond to the code blocks.
* Then write a separate function that turns a code block string into an exec.Cmd by doing substitutions, dealing with multi-line commands, etc

You could even have a separate data type for a code block (`type codeBlock string`) and then give it a method that returns the command.

tritone: Also I think this could have additional benefits as far as making testing easier, and allowing for support of other input formats other than README should that become necessary.

internal/lifecycle/readme.go
```diff
@@ -0,0 +1,205 @@
...
+// that code block.
+func codeBlocks(scanner *bufio.Scanner) ([][]string, error) {
+	var blocks [][]string
+
+	for scanner.Scan() {
+		line := scanner.Text()
+
+		if strings.Contains(line, codeTag) {
+			if s := scanner.Scan(); !s {
+				if err := scanner.Err(); err != nil && !s {
```
kurtisvg: nit: I think `!s` is already true if we're in this section?
```suggestion
				if err := scanner.Err(); err != nil {
```

internal/lifecycle/readme.go
```diff
@@ -0,0 +1,205 @@
...
+					return nil, fmt.Errorf("[lifecycle.codeBlocks] bufio.Scanner: %w", err)
+				}
+				return nil, fmt.Errorf("[lifecycle.codeBlocks]: unexpected EOF; file ended immediately " +
+					"after code tag")
```
kurtisvg: nit: Go _generally_ doesn't enforce a line limit, so it's typically fine to let a long line happen

If you really prefer to break the line, it's better to do something like this:
```go
return nil, fmt.Errorf(
       "[lifecycle.codeBlocks]: unexpected EOF; file ended immediately after code tag")
```
or this:
```go
s := fmt.Errorf("[lifecycle.codeBlocks]: unexpected EOF; file ended immediately after code tag")
return nil, s
```

The reason for this is that breaking in the middle of the string makes it harder to read - you're leaving it up to the user to mentally parse out the `" + "` and make sure everything lines up correctly.

internal/lifecycle/readme.go
```diff
@@ -0,0 +1,206 @@
...
+				line = line + l
+			}
+
+			line = os.ExpandEnv(line)
+			line = gcrURLRegexp.ReplaceAllString(line, gcrURL)
+			line = replaceServiceName(line, serviceName)
+			sp := strings.Split(line, " ")
+
+			var cmd *exec.Cmd
+			if strings.Contains(line, "gcloud") {
```
kurtisvg: Should this be "contains" or `sp[0] == "gcloud"`?

kurtisvg (changes requested):

internal/lifecycle/readme.go
SaketramDurbha: Resolved, thank you.

internal/lifecycle/readme.go
SaketramDurbha: Resolved, thank you

internal/lifecycle/readme.go
SaketramDurbha: Resolved, thank you

internal/lifecycle/readme.go
```diff
@@ -0,0 +1,201 @@
...
+	codeBlocks, err := codeBlocks(scanner)
+	if err != nil {
+		return nil, fmt.Errorf("[lifecycle.parseREADME] extracting code blocks out of %s: %w", filename, err)
+	}
+
+	if len(codeBlocks) == 0 {
+		return nil, errNoREADMECodeBlocksFound
+	}
+
+	lifecycle, err := codeBlocksTolifecycle(codeBlocks, serviceName, gcrURL)
```
tritone: I would define a custom type for codeBlock as a slice of strings to make it more clear what's going on. The surface would then look something like this (names might need a bit of thought...):

```
type codeBlock []string

func (cb codeBlock)  ToCommand(serviceName, gcrURL string) (*exec.Cmd, error) {...}

func extractBlocks(scanner *bufio.Scanner) ([]codeBlock, error) {...}
```

Then in parseREADME the logic would go (very basic sketch here):

```
blocks, err := extractBlocks(scanner)

// handle err

var l Lifecycle
for _, b := range blocks {
   c, err := b.toCommand(serviceName, gcrURL)
   // handle err
  l = append(l, c)
}

```

This allows you to re-use the code block parsing if there is some other source of input in the future, make it easier to test, and would better clarify what is happening where.

internal/lifecycle/readme.go
```diff
@@ -0,0 +1,201 @@
...
+			}
+
+			// If there is a backslash at the end of the line, this is a multiline command. Keep scanning to get
+			// entire command.
+			for line[len(line)-1] == bashLineContChar {
+				line = line[:len(line)-1]
+
+				i++
+				if i >= len(block) {
+					return nil, fmt.Errorf("[lifecycle.codeBlocksTolifecycle]: unexpected end of code block; expecting command line continuation")
```
tritone: no need for brackets around the function name in the error msg.

internal/sample/sample.go
```diff
@@ -66,7 +66,10 @@ func NewSample(dir string) (*Sample, error) {
 	}
 	service := gcloud.CloudRunService{Name: serviceName}
 
-	buildDeployLifecycle := lifecycle.NewLifecycle(dir, service.Name, cloudContainerImageURL)
+	buildDeployLifecycle, err := lifecycle.NewLifecycle(dir, service.Name, cloudContainerImageURL)
+	if err != nil {
+		return nil, fmt.Errorf("[sample.NewSample] getting Lifecycle: %w", err)
```
tritone, line 71: Again, no need for brackets

tritone (commented):
Nice improvement. I would also recommend adding tests to this PR if possible-- it will make it clearer what the right surface design is, and help you think about edge cases for your parsing logic.

internal/lifecycle/readme.go
SaketramDurbha: Thanks, planning to do this in a separate PR because other error messages outside of this PR are also formatted like this.

kurtisvg (approved):

internal/lifecycle/readme.go
SaketramDurbha: Thank you, resolved.

internal/lifecycle/readme.go
```diff
@@ -65,19 +118,24 @@ func parseREADME(filename, serviceName, gcrURL string) (Lifecycle, error) {
 		return nil, errNoREADMECodeBlocksFound
 	}
 
-	lifecycle, err := codeBlocksTolifecycle(codeBlocks, serviceName, gcrURL)
-	if err != nil {
-		return lifecycle, fmt.Errorf("[lifecycle.parseREADME] transforming code blocks in %s to executable commands: %w", filename, err)
+	var l Lifecycle
+	for _, b := range codeBlocks {
+		c, err := b.toCommands(serviceName, gcrURL)
```
tritone: Maybe rename `c` to something plural to indicate that it is a slice rather than a single command.

internal/lifecycle/readme.go
```diff
@@ -65,19 +118,24 @@ func parseREADME(filename, serviceName, gcrURL string) (Lifecycle, error) {
...
+		if err != nil {
+			return l, fmt.Errorf("[lifecycle.parseREADME] transforming code blocks in %s to executable commands: %w", filename, err)
+		}
+
+		l = append(l, c...)
```
tritone: Looking at this logic, I think that properly speaking you should redefine `NewLifecycle` to take a slice of codeBlock as a means of initializing, so you always have the same entry point. It breaks the encapsulation somewhat to have a NewLifecycle function but then have other ways of constructing a Lifecycle as well. The surface would be something like:

```
// Surfaces:
type LifecycleOption interface {
  apply(l *Lifecycle) // this method applies the option to the lifecycle.
}

type withCodeBlocks struct {
  cb []CodeBlock
}

func (w withCodeBlocks) apply(l *Lifecycle) {
  // this does the logic that you have in L122-129. 
}

func WithCodeBlocks(blocks []CodeBlock) LifecycleOption {
  // returns a withCodeBlocks struct using the given blocks.
}

// additional functions to produce appropriate lifecycle options as needed.

func NewLifecycle(opts ...LifecycleOption)  // This could have additional params such as service name if needed.

// To use:
l := NewLifecycle(FromCodeBlocks(blocks))
```

I don't think you have to fix this in order to merge this PR, but this is the kind of pattern that I would expect to see in order to polish this up.

tritone: Discussed this offline with Saketram-- I totally got the logic backwards here (since actually NewLifecycle is what calls the readme parsing logic), sorry about that!

I think it would be preferable for NewLifecycle to have options specifying what kind of lifecycle is desired rather than using defaults based on the environment or what files are present, but I'm not sure if that's fully compatible with project requirements.

README.md
SaketramDurbha: Resolved, thank you.

README.md
SaketramDurbha: Resolved, thank you.

internal/lifecycle/readme.go
SaketramDurbha: Resolved, thank you.

README.md
SaketramDurbha: README updated, thank you.

internal/lifecycle/readme.go
SaketramDurbha: Error messages updated, thank you.

internal/lifecycle/readme.go
SaketramDurbha: Irrelevant after refactor based on @tritone review.

internal/lifecycle/readme.go
SaketramDurbha: Yes -- added comment to clarify this, thank you.

internal/lifecycle/readme.go
SaketramDurbha: Resolved, thank you.

internal/lifecycle/readme.go
SaketramDurbha: Updated to check for return value of all `Scanner.Scan()` calls and check `Scanner.Err()` if appropriate.

internal/lifecycle/readme.go
SaketramDurbha: Mostly irrelevant after refactor based on @tritone review, thank you.

internal/lifecycle/lifecycle.go
SaketramDurbha: Resolved, thank you.

internal/lifecycle/readme.go
SaketramDurbha: Resolved, thank you.

internal/lifecycle/readme.go
SaketramDurbha: Same as here: https://github.com/GoogleCloudPlatform/serverless-sample-tester/pull/2#discussion_r467312902 . Thank you.

internal/lifecycle/readme.go
SaketramDurbha: Resolved, thank you.

internal/lifecycle/readme.go
SaketramDurbha: Resolved, thank you.

internal/lifecycle/lifecycle.go
SaketramDurbha: Resolved, thank you.

internal/lifecycle/lifecycle.go
SaketramDurbha: Resolved. Now we're only continuing to backup methods when either (a) the README.md file doesn't exist in the root directory or (b) it's obvious that the user hasn't tried to set the README up to work with our program (there are no tagged code blocks). This is per @kurtisvg suggestion in chat.

internal/lifecycle/lifecycle.go
SaketramDurbha: Resolved, thank you.

internal/lifecycle/readme.go
SaketramDurbha: Per offline discussion, we've decided that we'll keep pursuing it, but make sure to spend more time thinking about potential edge cases and keep an eye out for any that may come up. Also, will open an issue after this PR is merged for tests that will help us also help us catch edge cases.

internal/lifecycle/readme.go
SaketramDurbha: Resolved, thank you.

internal/lifecycle/readme.go
SaketramDurbha: Resolved, thank you.

internal/lifecycle/readme.go
SaketramDurbha: Resolved, thank you.

internal/lifecycle/readme.go
SaketramDurbha: Opening an issue for this after PR is merged.

internal/sample/sample.go
SaketramDurbha: Same as here: https://github.com/GoogleCloudPlatform/serverless-sample-tester/pull/2#discussion_r467152252 . Thank you.

internal/lifecycle/readme.go
SaketramDurbha: Resolved, thank you.

Issue #6, opened 2020-08-10
docs: README Clarity Improvements and Expanded Instructions


grayside:
This is some general feedback on the current state of the README as the project landing page and documentation.

1. In the introductory section, I do not understand what #3 is:

   > Return logs of health check service’s logs if any tests failed

2. Add setup steps before the usage example instead of the narrative below. For example:

   Authenticate gcloud with your user account:

   ```sh
   gcloud auth login
   ```

   Consider setting defaults for Cloud Run operations, such as setting the region:

   ```sh
   gcloud config set run/region us-central1`
   ```

   Run Serverless Sample Tester:

   ```sh
   ./sst [sample-dir]
   ```

3. The "README parsing" section dives quickly into technical details. I think a more general "Configuration & Implementation" section is needed, and should cover the following details:

   * All defaults provided by the tool
   * How commands can be customized
   * How tests can be customized

   As part of this, disamibiguate instructions for using the too from guidance on configuration overrides (e.g., I assume `platform/managed` is for customization, and not something the tool needs)

   Brief examples of scenarios overrides are useful to support would also be helpful. For example, in the case of Cloud Run,
   wanting to use the "gitops" mode of a service.yaml with `gcloud run services replace`.

This issue affects two categories in my review of the repository:
* **Adoption**: Can this project be found and does it provide enough information for developers to understand what it is, when to use it, and how to use it? Does it inspire confidence that it's a well-produced and well-maintained tool?
* **Experience**: Does the project provide a good experience to users?

PR #22, opened 2020-08-14
Config file and preliminary README changes

danielylee01:
Two things in this PR:
- Added (optional) config file functionality. Currently only contains customization for README file location 
- Preliminary changes to the README in response to #6. It does not comprehensively address everything in the issue, but it is a start on improving clarity

More to come later on both of these fronts.


README.md
```diff
@@ -59,9 +60,11 @@ supported.
...
-to your `gcloud run` commands.
+### README location
+If you wish to parse a README file located somewhere other than the root directory, you can include the README's location
+in a `config.yaml` file in the root directory, using the key `readme`.
 
-If comment code tags aren't added to your README, the program will fall back to reasonable defaults to build and deploy
-your sample to Cloud Run based on whether your sample is java-based and doesn't have a Dockerfile or isn't.
+For example:
+```text
+readme: ../README.md
```
kurtisvg, line 61: Is this path relative to the config.yaml? Relative to cwd? 

How does sst determine what the root directory is?

internal/lifecycle/lifecycle.go
```diff
@@ -47,10 +48,22 @@ func (l Lifecycle) Execute(commandsDir string) error {
 // those options are set up, it falls back to reasonable defaults based on whether the sample is java-based
 // (has a pom.xml) that doesn't have a Dockerfile or isn't.
 func NewLifecycle(sampleDir, serviceName, gcrURL string) (Lifecycle, error) {
-	readmePath := filepath.Join(sampleDir, "README.md")
+	var readmePath string
+	// Searching for config file
+	if err := viper.ReadInConfig(); err == nil {
```
kurtisvg, line 53: At some point (doesn't need to be in this PR) you should probably consolidate all of the configuration and parse logic into a single place (probably in the `cmd` module). Otherwise as complexity grows, you'll be reimplementing and tracking the correct order of operations for configuration conflicts in multiple different locations in your code base.

kurtisvg (changes requested):
You mentioned feedback from Adam - does this PR fix or address an open issue? if so, you should link to in the body of this PR

README.md
danielylee01: This path is relative to the root directory of the sample directory specified on the command line; I'll update the README so this is more clear. (It will also work if the user specifies an absolute directory.) If you think a different path would make more sense, I can change it.

kurtisvg: As is, the README still isn't clear. It says "root directory", but doesn't define what that is. I think what you really men to say is _target_ directory - it's relative to the directory you passed to sst (which is also where the `config.yaml` can be found)

kurtisvg (changes requested):

danielylee01:
Refined README in response to feedback from Kurtis, and continued addressing suggestions in #6.

kurtisvg (approved):

PR #23, opened 2020-08-14
refactor error messages (fixes #4)

### LINKED ISSUE #4: Error message formatting
Error messages are currently formatted like this:
```
if err != nil {
    fmt.Errorf("[{package/struct name}.{function name}] {what we were trying to do}: %v", err)
}
```

Per @tritone, we can drop the `[ ]` around the package/struct name and function name:
```
if err != nil {
    fmt.Errorf("{package/struct name}.{function name} {what we were trying to do}: %v", err)
}
```

SaketramDurbha:
This PR fixes issue #4.

- Include callee function rather than caller functions
- Add more error context in some cases

Update error message format from:
```
if err != nil {
    fmt.Errorf("[{package/struct name}.{current function name}] {what we were trying to do}: %w", err)
}
```

To:
```
if err != nil {
    fmt.Errorf("{package/struct name}.{callee function name} {what we were trying to do}: %w", err)
}
```

### DIFF
--- a/internal/cmd/cmd.go
+++ b/internal/cmd/cmd.go
@@ -51,7 +51,7 @@ func Root(cmd *cobra.Command, args []string) error {
 	defer s.Service.Delete(s.Dir)
 	defer s.DeleteCloudContainerImage()
 	if err != nil {
-		return fmt.Errorf("[cmd.Root] building and deploying sample to Cloud Run: %w", err)
+		return fmt.Errorf("Lifecycle.Execute building and deploying sample to Cloud Run: %w", err)
 	}
 
 	log.Println("Getting identity token for gcloud auhtorized account")
@@ -61,19 +61,19 @@ func Root(cmd *cobra.Command, args []string) error {
 	identToken, err = util.ExecCommand(exec.Command("gcloud", a...), s.Dir)
 
 	if err != nil {
-		return fmt.Errorf("[cmd.Root] getting identity token for gcloud auhtorized account: %w", err)
+		return fmt.Errorf("util.ExecCommand getting identity token for gcloud auhtorized account: %w", err)
 	}
 
 	log.Println("Checking endpoints for expected results")
 	serviceURL, err := s.Service.URL(s.Dir)
 	if err != nil {
-		return fmt.Errorf("[cmd.Root] getting Cloud Run service URL: %w", err)
+		return fmt.Errorf("Service.URL %s dir: %w", s.Dir, err)
 	}
 
 	log.Println("Validating Cloud Run service endpoints for expected status codes")
 	allTestsPassed, err := util.ValidateEndpoints(serviceURL, &swagger.Paths, identToken)
 	if err != nil {
-		return fmt.Errorf("[cmd.Root] validating Cloud Run service endpoints for expected status codes: %w", err)
+		return fmt.Errorf("util.ValidateEndpoints: %w", err)
 	}
 
 	if !allTestsPassed {
--- a/internal/cmd/cmd.go
+++ b/internal/cmd/cmd.go
@@ -51,7 +51,7 @@ func Root(cmd *cobra.Command, args []string) error {
 	defer s.Service.Delete(s.Dir)
 	defer s.DeleteCloudContainerImage()
 	if err != nil {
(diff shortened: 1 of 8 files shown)


kurtisvg (approved):

internal/lifecycle/readme.go
```diff
@@ -143,15 +143,15 @@ func extractCodeBlocks(scanner *bufio.Scanner) ([]codeBlock, error) {
...
 				}
-				return nil, fmt.Errorf("[lifecycle.codeBlocks]: unexpected EOF; file ended immediately after code tag")
+				return nil, fmt.Errorf("unexpected EOF: file ended immediately after code tag")
 			}
 
 			startCodeBlockLine := scanner.Text()
 			m := mdCodeFenceStartRegexp.MatchString(startCodeBlockLine)
 			if !m {
-				return nil, fmt.Errorf("[lifecycle.codeBlocks]: expecting start of code block immediately after code tag")
+				return nil, fmt.Errorf("expecting start of code block immediately after code tag")
```
tritone: These errors might benefit from adding which line of parsing they failed on; I think they'd be confusing otherwise. Can be accessed via https://golang.org/pkg/text/scanner/#Scanner.Pos looks like.

internal/util/endpoints.go
```diff
@@ -109,21 +109,21 @@ func makeTestRequest(endpointURL, httpMethod, mimeType string, reqBodyReader *st
 
 	req, err := http.NewRequest(httpMethod, endpointURL, reqBodyReader)
 	if err != nil {
-		return false, fmt.Errorf("[util.makeTestRequest] creating an http.Request: %w", err)
+		return false, fmt.Errorf("http.NewRequest creating an http.Request: %w", err)
```
tritone: Message here seems redundant.

internal/gcloud/cloud_run_service.go
```diff
@@ -41,7 +41,7 @@ func (s CloudRunService) Delete(sampleDir string) error {
 	_, err := util.ExecCommand(exec.Command("gcloud", a...), sampleDir)
 
 	if err != nil {
-		return fmt.Errorf("[CloudRunService.delete] deleting Cloud Run Service: %w", err)
+		return fmt.Errorf("util.ExecCommand deleting Cloud Run Service: %w", err)
```
tritone: I would say that these and the other util.ExecCommand errors don't need to include the method name. The error messages that you have are what's most useful here and ExecCommand is just a thin wrapper. There's some room for debate here probably but that's my inclination.

tritone (commented):
Overall looks good, a couple thoughts on error messages.

internal/cmd/cmd.go
```diff
@@ -46,7 +46,7 @@ func Root(cmd *cobra.Command, args []string) error {
 	defer s.Service.Delete(s.Dir)
 	defer s.DeleteCloudContainerImage()
 	if err != nil {
-		return fmt.Errorf("[cmd.Root] building and deploying sample to Cloud Run: %w", err)
+		return fmt.Errorf("Lifecycle.Execute building and deploying sample to Cloud Run: %w", err)
```
grayside: nit: Usual format is "package.Function: [optional message]: %w", note the colon after the method name.

internal/gcloud/cloud_run_service.go
grayside: Agreed given there's not a lot of different ways to exec in this project.

I often find it is useful in programs which passthru and execute shell commands to actually log the command used, especially in an error. However, that is a further logging enhancement out of scope of this PR.

internal/cmd/cmd.go
SaketramDurbha: Resolved, thank you.

internal/gcloud/cloud_run_service.go
SaketramDurbha: Resolved, thank you.

internal/util/endpoints.go
SaketramDurbha: Resolved, thank you.

internal/lifecycle/readme.go
SaketramDurbha: Resolved offline, thank you. We're printing the line number during `extractCodeBlocks` and dumping the entire code block elsewhere.

tritone (approved):

PR #24, opened 2020-08-14
run README build and deploy commands through bash shell (fix #21)

### LINKED ISSUE #21: Shell for running README parsing build and deplpy commands
We originally didn't pass build and deploy commands for build and deploy commands parsed out of Markdown files through a shell was because Go's `exec.Cmd` interface doesn't automatically do it for us. But just recently found out that we can do something like this:
```
exec.Command("bash", "-c", "echo hello world")
```

And we could revise the code tag to include the shell we want to run the commands with instead: `{sst-run-bash}` instead of `{sst-run-bash}`. This way, we wouldn't have to expand environment variables ourselves either.

SaketramDurbha:
Runs build and deploy commands through the Bash shell. 

Support for more shells to come.

Will rebase and update once #22 is merged.

### DIFF
--- a/README.md
+++ b/README.md
@@ -49,7 +49,7 @@ inside a code fence that is immediately preceded by a line containing `{sst-run-
 (e.g. `[//]: # ({sst-run-bash})`) to include this line without making it visible when rendered.
 
 The parsed commands will be run through the Bash shell, but each command will be run through a separate instance
-of the shell, meaning that commands such as `cd` may not work as expected.  
+of the shell, meaning that commands such as `cd` or `export` may not work as expected.  
 
 However, any environment variables referenced in the form of `$var` or `${var}` will expanded. In addition, bash-style
 multiline commands (i.e. non-quoted backslashes at the end of a line that indicate a line continuation) will also be 
--- a/README.md
+++ b/README.md
@@ -49,11 +49,15 @@ inside a code fence that is immediately preceded by a line containing `{sst-run-
 (e.g. `[//]: # ({sst-run-bash})`) to include this line without making it visible when rendered.
 
 The parsed commands will be run through the Bash shell, but each command will be run through a separate instance
-of the shell, meaning that commands such as `cd` or `export` may not work as expected.  
+of the shell, meaning that commands such as `cd` or `export` may not work as expected. All commands will be run in the
+sample's root directory.
 
-However, any environment variables referenced in the form of `$var` or `${var}` will expanded. In addition, bash-style
-multiline commands (i.e. non-quoted backslashes at the end of a line that indicate a line continuation) will also be 
-supported. 
+However, any environment variables referenced in the form of `$var` or `${var}` will expanded. Each command's process
+will share the same environment as the process you execute `sst` from. So environment variables you set in your shell
+before calling `sst` will be available for the commands you specify in your README. 
+
+In addition, bash-style multiline commands (i.e. non-quoted backslashes at the end of a line that indicate a line
+continuation) will also be supported.
 
 Do not set the Cloud Run region you'd like to deploy to through the `--region` flag in the `gcloud run` commands.
 Instead, as mentioned above, do so by setting the `run/region` gcloud property.
--- a/README.md
+++ b/README.md
@@ -33,24 +33,23 @@ If you'd like, make sure to include the following comment code tag in your READM
 build and deploy your sample:
 
 ```text
-[//]: # ({sst-run-unix})
+[//]: # ({sst-run-bash})
 ```
 
 For example:
 ````text
-[//]: # ({sst-run-unix})
+[//]: # ({sst-run-bash})
 ```
 gcloud builds submit --tag=gcr.io/${GOOGLE_CLOUD_PROJECT}/run-mysql
 ```
 ````
 
 When parsing the README for custom build and deploy commands, the serverless sample tester will include any commands
-inside a code fence that is immediately preceded by a line containing `{sst-run-unix}`. You can use Markdown syntax
-(e.g. `[//]: # ({sst-run-unix})`) to include this line without making it visible when rendered.
+inside a code fence that is immediately preceded by a line containing `{sst-run-bash}`. You can use Markdown syntax
+(e.g. `[//]: # ({sst-run-bash})`) to include this line without making it visible when rendered.
 
-The parsed commands will not be run through a shell, meaning that the program will not perform any expansions,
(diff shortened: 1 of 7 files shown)


SaketramDurbha:
This PR fixes issue #21, by the way.

internal/lifecycle/readme.go
```diff
@@ -79,17 +79,17 @@ func (cb codeBlock) toCommands(serviceName, gcrURL string) ([]*exec.Cmd, error)
...
 		line = replaceServiceName(line, serviceName)
-		sp := strings.Split(line, " ")
 
 		var cmd *exec.Cmd
-		if sp[0] == "gcloud" {
-			a := append(util.GcloudCommonFlags, sp[1:]...)
-			cmd = exec.Command("gcloud", a...)
+		if gcloudCommandRegexp.MatchString(line) {
+			sp := strings.Split(line, " ")
+			c := "gcloud " + strings.Join(append(util.GcloudCommonFlags, sp[1:]...), " ")
```
kurtisvg: I think you may want to avoid strings.Join here, and instead just pass it as arguments into exec.Command and let it concat it incase anything special needs to be done on different OSs

internal/lifecycle/readme.go
```diff
@@ -79,17 +79,17 @@ func (cb codeBlock) toCommands(serviceName, gcrURL string) ([]*exec.Cmd, error)
...
 		line = gcrURLRegexp.ReplaceAllString(line, gcrURL)
 		line = replaceServiceName(line, serviceName)
-		sp := strings.Split(line, " ")
 
 		var cmd *exec.Cmd
-		if sp[0] == "gcloud" {
-			a := append(util.GcloudCommonFlags, sp[1:]...)
-			cmd = exec.Command("gcloud", a...)
+		if gcloudCommandRegexp.MatchString(line) {
+			sp := strings.Split(line, " ")
```
kurtisvg: This might be an issue in the future - different systems have different split rules, so you may want to find a library that splits the same way bash does. You can put a TODO here for now and open an issue to track if you want.

kurtisvg (changes requested):

internal/lifecycle/readme.go
SaketramDurbha: Thank you for your comment.

The reason we joined the string back up was because I believe `bash` expects the entire command as a single argument. Testing this on the command line, `bash -c echo $HELLO_WORLD` didn't print anything whereas `bash -c "echo $HELLO_WORLD"` printed the expected value. Similarly, testing this in Go,the output of `exec.Command("bash", "-c", "echo", "$HELLO_WORLD")` was nothing and output of `exec.Command("bash", "-c", "echo $HELLO_WORLD")` was as expected.

This is part of the reason we hadn't implemented support for other shells yet, since we're not sure which executable and flags we need to use for running a command through the shell and how that executable expects the command to be passed to it.

internal/lifecycle/readme.go
SaketramDurbha: Okay, thank you, opening an issue for this after this PR gets merged and adding a comment in the code.

kurtisvg (approved):

README.md
```diff
@@ -33,24 +33,23 @@ If you'd like, make sure to include the following comment code tag in your READM
...
-inside a code fence that is immediately preceded by a line containing `{sst-run-unix}`. You can use Markdown syntax
-(e.g. `[//]: # ({sst-run-unix})`) to include this line without making it visible when rendered.
+inside a code fence that is immediately preceded by a line containing `{sst-run-bash}`. You can use Markdown syntax
+(e.g. `[//]: # ({sst-run-bash})`) to include this line without making it visible when rendered.
 
-The parsed commands will not be run through a shell, meaning that the program will not perform any expansions,
-pipelines, redirections or any other functions that shells are responsible for. This also means that popular shell
-builtin commands like `cd`, `export`, and `echo` will not be available or may not work as expected.  
+The parsed commands will be run through the Bash shell, but each command will be run through a separate instance
+of the shell, meaning that commands such as `cd` or `export` may not work as expected.
```
tritone: Can you clarify how users are supposed to do control dir of execution or env variables, then? Can I chain commands via `&&` for example?

internal/lifecycle/readme.go
```diff
@@ -79,17 +79,19 @@ func (cb codeBlock) toCommands(serviceName, gcrURL string) ([]*exec.Cmd, error)
...
-		sp := strings.Split(line, " ")
 
 		var cmd *exec.Cmd
-		if sp[0] == "gcloud" {
-			a := append(util.GcloudCommonFlags, sp[1:]...)
-			cmd = exec.Command("gcloud", a...)
+		if gcloudCommandRegexp.MatchString(line) {
+			// Splitting command to add util.GcloudCommonFlags to front of command.
+			// TODO: Ideally want to use a library that splits commands according to bash specification.
+			sp := strings.Split(line, " ")
```
tritone: Could you just split once (after `gcloud`) using `SplitN`? I'm concerned that this will alter whitespace later in the command, which could be an issue (e.g. for whitespace in a quoted arg).

grayside: I strongly recommend avoiding the game of CLI arg manipulation. Most gcloud flags are also available with config and environment variable overrides. I think the latter would work nicely here.

Override the environment for the `exec.Command` operation by combining the existing environment variables with the items you need to customize.

The Go technique: https://pkg.go.dev/os/exec?tab=doc#example-Command-Environment
For `--quiet`, the documented variable is `CLOUDSDK_CORE_DISABLE_PROMPTS`.

SaketramDurbha: Resolved, thank you both! We're now overriding the environment using the method outlined by the webpage @grayside linked to above.

README.md
SaketramDurbha: Thank you for your comment.

For environment variables, each command's process will uses the environment of the process from which users execute our program. So environment variables they have set in their shell where they call `sst` will be available for the commands that they write in their README. I'm going to adding a comment in our README to clarify this.

For the directory of execution, we're not currently supporting the option for users to control the directory where their README commands are executed; all commands will be executed in the sample's root directory (I going to a comment in our README to clarify this as well). But we're going to open an issue for this feature to make sure that it's tracked.

However, as you mentioned, you are able to get around this by chaining commands together, like `cd .. && ls`. But 
I think we're not going to put this in our documentation because we believe a more permanent solution is preferred, and since each command is run in it's own separate instance of a shell, so you would have to chain each where you want a directory command with a `cd`.

Hope this makes sense. Did this answer your questions?

PR #25, opened 2020-08-14
add readme parsing tests

### LINKED ISSUE #5: README parsing tests
This project is currently lacking tests, including for complex parts of the code like README parsing.

Tests should be implemented to primarily ensure that edge cases for parsing build and deploy commands out of Markdown files are covered.

SaketramDurbha:
This PR addresses issue #5 .

- Add readme parsing tests
- Refactor acutal readme parsing code to be more testable

### DIFF
--- a/internal/lifecycle/lifecycle.go
+++ b/internal/lifecycle/lifecycle.go
@@ -59,7 +59,7 @@ func NewLifecycle(sampleDir, serviceName, gcrURL string) (Lifecycle, error) {
 	}
 
 	if _, err := os.Stat(readmePath); err == nil {
-		lifecycle, err := parseReadme(readmePath, serviceName, gcrURL)
+		lifecycle, err := parseREADME(readmePath, serviceName, gcrURL)
 		// Show README location
 		log.Println("README.md location: " + readmePath)
 		if err == nil {
@@ -68,7 +68,7 @@ func NewLifecycle(sampleDir, serviceName, gcrURL string) (Lifecycle, error) {
 		}
 
 		if !errors.Is(err, errNoReadmeCodeBlocksFound) {
-			return nil, fmt.Errorf("lifecycle.parseReadme: %s: %w", readmePath, err)
+			return nil, fmt.Errorf("lifecycle.parseREADME: %s: %w", readmePath, err)
 		}
 
 		log.Printf("No code blocks immediately preceded by %s found in README.md\n", codeTag)
--- a/internal/lifecycle/lifecycle.go
+++ b/internal/lifecycle/lifecycle.go
@@ -73,7 +73,7 @@ func NewLifecycle(sampleDir, serviceName, gcrURL string) (Lifecycle, error) {
 			return nil, fmt.Errorf("lifecycle.parseREADME: %s: %w", readmePath, err)
 		}
 
-		log.Println("No code blocks immediately preceded by %s found in README.md\n", codeTag)
+		log.Printf("No code blocks immediately preceded by %s found in README.md\n", codeTag)
 	} else {
 		log.Println("No README.md found")
 	}
--- a/internal/lifecycle/lifecycle.go
+++ b/internal/lifecycle/lifecycle.go
@@ -18,11 +18,11 @@ import (
 	"errors"
 	"fmt"
 	"github.com/GoogleCloudPlatform/serverless-sample-tester/internal/util"
+	"github.com/spf13/viper"
 	"log"
 	"os"
 	"os/exec"
 	"path/filepath"
-	"github.com/spf13/viper"
 )
 
 // Lifecycle is a list of ordered exec.Cmd that should be run to execute a certain process.
@@ -58,10 +58,8 @@ func NewLifecycle(sampleDir, serviceName, gcrURL string) (Lifecycle, error) {
 		readmePath = filepath.Join(sampleDir, "README.md")
 	}
 
-
-
 	if _, err := os.Stat(readmePath); err == nil {
-		lifecycle, err := parseREADME(readmePath, serviceName, gcrURL)
+		lifecycle, err := parseReadme(readmePath, serviceName, gcrURL)
 		// Show README location
 		log.Println("README.md location: " + readmePath)
 		if err == nil {
@@ -69,8 +67,8 @@ func NewLifecycle(sampleDir, serviceName, gcrURL string) (Lifecycle, error) {
 			return lifecycle, nil
 		}
 
-		if !errors.Is(err, errNoREADMECodeBlocksFound) {
-			return nil, fmt.Errorf("lifecycle.parseREADME: %s: %w", readmePath, err)
+		if !errors.Is(err, errNoReadmeCodeBlocksFound) {
+			return nil, fmt.Errorf("lifecycle.parseReadme: %s: %w", readmePath, err)
 		}
 
 		log.Printf("No code blocks immediately preceded by %s found in README.md\n", codeTag)
--- a/internal/lifecycle/readme.go
+++ b/internal/lifecycle/readme.go
@@ -102,9 +102,10 @@ func (cb codeBlock) toCommands(serviceName, gcrURL string) ([]*exec.Cmd, error) 
 	return cmds, nil
 }
 
-// parseREADME parses a README file with the given name. It reads terminal commands surrounded by one of the codeTags
-// listed above and loads them into a Lifecycle. In the process, it replaces the Cloud Run service name and Container
-// Registry tag with the provided inputs.
+// parseREADME parses a README file with the given name. It parses terminal commands in code blocks annotated by the
+// codeTag and loads them into a Lifecycle. In the process, it replaces the Cloud Run service name and Container
+// Registry tag with the provided inputs. It also expands environment variables and supports bash-style line
+// continuations.
 func parseREADME(filename, serviceName, gcrURL string) (Lifecycle, error) {
 	file, err := os.Open(filename)
 	if err != nil {
@@ -114,14 +115,13 @@ func parseREADME(filename, serviceName, gcrURL string) (Lifecycle, error) {
 
 	scanner := bufio.NewScanner(file)
 
-	l, err := extractLifecycle(scanner, serviceName, gcrURL)
-	if err != nil {
-		return l, err
-	}
-
-	return l, nil
+	return extractLifecycle(scanner, serviceName, gcrURL)
 }
 
+// extractLifecycle is a helper function for parseREADME. It takes a scanner that reads from a Markdown file and parses
+// terminal commands in code blocks annotated by the codeTag and loads them into a Lifecycle. In the process, it
+// replaces the Cloud Run service name and Container Registry tag with the provided inputs. It also expands environment
+// variables and supports bash-style line continuations.
 func extractLifecycle(scanner *bufio.Scanner, serviceName, gcrURL string) (Lifecycle, error) {
 	codeBlocks, err := extractCodeBlocks(scanner)
 	if err != nil {
--- a/internal/lifecycle/readme.go
+++ b/internal/lifecycle/readme.go
@@ -109,9 +109,19 @@ func parseREADME(filename, serviceName, gcrURL string) (Lifecycle, error) {
 	defer file.Close()
 
 	scanner := bufio.NewScanner(file)
+
+	l, err := extractLifecycle(scanner, serviceName, gcrURL)
+	if err != nil {
+		return l, fmt.Errorf("[lifecycle.parseREADME] extracting commands out of %s: %w", filename, err)
+	}
+	return l, nil
+
+}
+
+func extractLifecycle(scanner *bufio.Scanner, serviceName, gcrURL string) (Lifecycle, error) {
 	codeBlocks, err := extractCodeBlocks(scanner)
 	if err != nil {
-		return nil, fmt.Errorf("lifecycle.extractCodeBlocks: %s: %w", filename, err)
+		return nil, fmt.Errorf("lifecycle.extractCodeBlocks: %w", err)
 	}
 
 	if len(codeBlocks) == 0 {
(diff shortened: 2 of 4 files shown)


internal/lifecycle/readme.go
```diff
@@ -42,7 +42,11 @@ var (
 
 	mdCodeFenceStartRegexp = regexp.MustCompile("^\\w*`{3,}[^`]*$")
 
-	errNoREADMECodeBlocksFound = fmt.Errorf("[lifecycle.parseREADME]: no code blocks immediately preceded by %s found", codeTag)
+	errNoREADMECodeBlocksFound   = fmt.Errorf("[lifecycle.parseREADME]: no code blocks immediately preceded by %s found", codeTag)
```
kurtisvg: We drop the `[funcName]` in another PR I think - should we do so here as well?

kurtisvg: if these errors aren't being returned in multiple locations, It's probably not worth declaring in advance.

internal/lifecycle/readme.go
```diff
@@ -109,9 +113,19 @@ func parseREADME(filename, serviceName, gcrURL string) (Lifecycle, error) {
...
+
+	l, err := extractLifecycle(scanner, serviceName, gcrURL)
+	if err != nil {
+		return l, err
+	}
+
+	return l, nil
+}
+
+func extractLifecycle(scanner *bufio.Scanner, serviceName, gcrURL string) (Lifecycle, error) {
```
kurtisvg, line 125: Should you add a comment to this function?

internal/lifecycle/readme_test.go
```diff
@@ -0,0 +1,438 @@
...
+)
+
+func setEnv(e map[string]string) error {
+	for k, v := range e {
+		err := os.Setenv(k, v)
+		if err != nil {
+			return err
+		}
+	}
+
```
kurtisvg: nit: remove newline

internal/lifecycle/readme_test.go
```diff
@@ -0,0 +1,438 @@
...
+	"strings"
+	"testing"
+)
+
+func setEnv(e map[string]string) error {
+	for k, v := range e {
+		err := os.Setenv(k, v)
+		if err != nil {
+			return err
+		}
```
kurtisvg: nit:
```suggestion
		if err := os.Setenv(k, v); err != nil {
			return err
		}
```

internal/lifecycle/readme_test.go
```diff
@@ -0,0 +1,438 @@
...
+	"bufio"
+	"errors"
+	"os"
+	"os/exec"
+	"reflect"
+	"strings"
+	"testing"
+)
+
+func setEnv(e map[string]string) error {
```
kurtisvg, line 14: nit: Comments are good, even on utility functions like this.

internal/lifecycle/readme_test.go
```diff
@@ -0,0 +1,438 @@
...
+}
+
+func equalError(a, b error) bool {
+	if a == nil {
+		return b == nil
+	}
+	if b == nil {
+		return a == nil
+	}
+	return errors.Is(a, b)
```
kurtisvg: hmm, I'm not sure what this is doing, but it seems iffy. What are you trying to do?

internal/lifecycle/readme_test.go
```diff
@@ -0,0 +1,438 @@
...
+	extractLifecycleErr  error
+	extractCodeBlocksErr error
+	env                  map[string]string
+	serviceName          string
+	gcrURL               string
+}
+
+var tests = []test{
+	{
+		in: "[//]: # ({sst-run-unix})\n" +
```
kurtisvg: Consider multiline strings or a single line - this feels a bit hard to read.

internal/lifecycle/readme_test.go
```diff
@@ -0,0 +1,438 @@
...
+	}
+}
+
+func TestExtractCodeBlocks(t *testing.T) {
+	for i, tc := range tests {
+		s := bufio.NewScanner(strings.NewReader(tc.in))
+
+		h, err := extractCodeBlocks(s)
+		if !equalError(err, tc.extractCodeBlocksErr) {
+			t.Errorf("#%d: error mismatch\nhave: %v\nwant: %v", i, err, tc.extractCodeBlocksErr)
```
kurtisvg: Go has a common idiom for this: `want x, got y`

internal/lifecycle/readme_test.go
```diff
@@ -0,0 +1,438 @@
...
+	toCommandsErr        error
+	extractLifecycleErr  error
+	extractCodeBlocksErr error
+	env                  map[string]string
+	serviceName          string
+	gcrURL               string
+}
+
+var tests = []test{
+	{
```
kurtisvg, line 49: I would consider adding a comment to each of these - what are these parameters supposed to be testing?

kurtisvg (changes requested):

internal/lifecycle/readme_test.go
SaketramDurbha: Resolved, thank you.

internal/lifecycle/readme_test.go
SaketramDurbha: Resolved, thank you.

internal/lifecycle/readme_test.go
SaketramDurbha: Resolved, thank you.

internal/lifecycle/readme_test.go
SaketramDurbha: Resolved, thank you.

internal/lifecycle/readme.go
SaketramDurbha: Resolved, thank you.

internal/lifecycle/readme_test.go
SaketramDurbha: Resolved, thank you.

internal/lifecycle/readme_test.go
SaketramDurbha: Resolved offline, thank you.

internal/lifecycle/readme.go
SaketramDurbha: Resolved offline, thank you.

internal/lifecycle/readme_test.go
SaketramDurbha: Resolved offline, thank you.

internal/lifecycle/readme_test.go
SaketramDurbha: Resolved offline, thank you.

internal/lifecycle/readme_test.go
SaketramDurbha: For reference: we're keeping this because it doesn't seem line multiline strings support escaped backticks.

kurtisvg (approved):

internal/lifecycle/readme_test.go
```diff
@@ -0,0 +1,504 @@
...
+		cmds: []*exec.Cmd{
+			exec.Command("echo", "hello", "world"),
+			exec.Command("echo", "line", "one"),
+			exec.Command("echo", "line", "two"),
+		},
+	},
+}
+
+func TestToCommands(t *testing.T) {
+	for i, tc := range tests {
```
kurtisvg: Rather than iterate through each list of tests for each section, you should probably just have a different slice for each test.

kurtisvg (approved):

internal/lifecycle/readme_test.go
SaketramDurbha: Resolved, thank you.

internal/lifecycle/readme_test.go
```diff
@@ -0,0 +1,448 @@
...
+			t.Errorf("#%d: result mismatch\nwant: %#+v\ngot: %#+v", i, tc.cmds, cmds)
+		}
+
+		if err := unsetEnv(tc.env); err != nil {
+			t.Errorf("#%d: unsetEnv: %v", i, err)
+		}
+	}
+}
+
+type parseREADMETest struct {
```
kurtisvg, line 233: nit: Use MixedCaps

internal/lifecycle/readme_test.go
```diff
@@ -0,0 +1,448 @@
...
+
+		// Cloud Run Service name and Container Registry URL tag replacement will be tested in TestToCommands
+		lifecycle, err := extractLifecycle(s, "", "")
+
+		if !errors.Is(err, tc.err) {
+			t.Errorf("#%d: error mismatch\nwant: %v\ngot: %v", i, tc.err, err)
+			continue
+		}
+
+		if err == nil && !reflect.DeepEqual(lifecycle, tc.lifecycle) {
```
kurtisvg, line 324: What if err != nil?

kurtisvg (approved):
2 nits, then LGTM

internal/lifecycle/readme_test.go
SaketramDurbha: I wasn't sure if we wanted to still check the results if there was an error.

internal/lifecycle/readme_test.go
SaketramDurbha: Resolved, thank you.

internal/lifecycle/readme_test.go
SaketramDurbha: After talking offline, we're leaving this as is. Thank you.

dmahugh (approved):
LGTM

internal/lifecycle/lifecycle.go
```diff
@@ -58,22 +58,20 @@ func NewLifecycle(sampleDir, serviceName, gcrURL string) (Lifecycle, error) {
 		readmePath = filepath.Join(sampleDir, "README.md")
 	}
 
-
-
 	if _, err := os.Stat(readmePath); err == nil {
-		lifecycle, err := parseREADME(readmePath, serviceName, gcrURL)
```
tritone, line 64: All caps on README is actually appropriate here I think.

internal/lifecycle/readme.go
```diff
@@ -42,7 +42,11 @@ var (
 
 	mdCodeFenceStartRegexp = regexp.MustCompile("^\\w*`{3,}[^`]*$")
 
-	errNoREADMECodeBlocksFound = fmt.Errorf("lifecycle.extractCodeBlocks: no code blocks immediately preceded by %s found", codeTag)
+	errNoReadmeCodeBlocksFound   = fmt.Errorf("lifecycle.extractCodeBlocks: no code blocks immediately preceded by %s found", codeTag)
```
tritone, line 45: I actually don't think this change is necessary, unless any of these errors are going to be used in more than one place. It makes it a little more complicated to read the code.

If anything, it might be easier for callers if there was a single error type that wrapped any error coming from readme parsing.

internal/lifecycle/readme.go
```diff
@@ -68,7 +72,7 @@ func (cb codeBlock) toCommands(serviceName, gcrURL string) ([]*exec.Cmd, error)
 
 			i++
 			if i >= len(cb) {
-				return nil, fmt.Errorf("unexpected end of code block: expecting command line continuation; code block dump:\n%s", strings.Join(cb, "\n"))
+				return nil, fmt.Errorf("%w; code block dump:\n%s", errCodeBlockEndAfterLineCont, strings.Join(cb, "\n"))
```
tritone: The use of wrapping here doesn't really make sense-- usually that would just be when you are bubbling up an error from a called func.

internal/lifecycle/readme.go
```diff
@@ -204,15 +218,15 @@ func replaceServiceName(command, serviceName string) string {
...
+	for i := 0; i < len(sp); i++ {
 		if sp[i] == os.ExpandEnv("$CLOUD_RUN_SERVICE_NAME") {
 			sp[i] = serviceName
 			return strings.Join(sp, " ")
 		}
 	}
 
 	// Searches for specific gcloud keywords and takes service name from them
-	for i := 0; i < len(sp) - 1; i++ {
+	for i := 0; i < len(sp)-1; i++ {
```
tritone, line 229: Were these formatting changes yours or from gofmt?

internal/lifecycle/readme_test.go
```diff
@@ -0,0 +1,452 @@
...
+			exec.Command("gcloud", "--quiet", "run", "services", "update", uniqueServiceName, "--add-cloudsql-instances=project:region:instance"),
+		},
+		env: map[string]string{
+			"TEST_CLOUD_SQL_CONNECTION": "project:region:instance",
+		},
+	},
+}
+
+func TestToCommands(t *testing.T) {
+	for i, tc := range toCommandsTests {
```
tritone, line 195: Having a list of test cases and then looping through them seems like the right approach. I would also consider using a subtest for each case, which produces nicer output. This seems like it could work for most of your tests.

https://golang.org/pkg/testing/#hdr-Subtests_and_Sub_benchmarks
https://blog.golang.org/subtests

internal/lifecycle/readme_test.go
```diff
@@ -0,0 +1,452 @@
...
+		},
+	},
+
+	// line cont char but code block closes at next line
+	{
+		codeBlock: codeBlock{
+			"echo multi \\",
+		},
+		cmds: nil,
+		err:  errCodeBlockEndAfterLineCont,
```
tritone, line 87: In general it's better not to have your test use unexported stuff from the package. Think about what kind of error a caller outside the package would expect to see. String matching is probably fine, or you could think about making an exported error that callers could actually check for.

SaketramDurbha: Resolved offline, thank you.

internal/lifecycle/readme.go
SaketramDurbha: Resolved offline, thank you.

internal/lifecycle/readme.go
SaketramDurbha: `gofmt`.

internal/lifecycle/readme_test.go
SaketramDurbha: Creating an issue for this.

internal/lifecycle/lifecycle.go
SaketramDurbha: Resolved, thank you.

internal/lifecycle/readme.go
SaketramDurbha: Resolved, thank you.

internal/lifecycle/readme_test.go
SaketramDurbha: Actually, a PR: https://github.com/GoogleCloudPlatform/serverless-sample-tester/pull/31

Issue #17, opened 2020-08-10
Add support for configurable timeouts


grayside:
### Current Behavior

#16 suggests setting an explicit default timeout of 10 seconds per request.
However, different sample tests may take different lengths of time, either because the sample takes more time to process request, the number of test cases to be verified is longer, or both.

### Proposed Solution

Add a `--timeout` flag.

---

This issue affects two categories in my review of the repository:
* **Experience**: Does the project provide a good experience to users?
* **Scope**: Does the tool have the breadth of features and flexibility to support reasonable testing needs?

PR #26, opened 2020-08-14
add timeout of 10s to HTTP requests made to cloud run services (fix #16)

### LINKED ISSUE #16: Fix unbounded timeout for test HTTP requests
### Current Behavior

"Endpoint Validation" HTTP requests have no timeout. As such, certain types of test failures could lead to tests being stuck for minutes or hours.

### Proposed Solution

Instead of using `http.DefaultClient` in [`makeTestRequest()`](https://github.com/GoogleCloudPlatform/serverless-sample-tester/blob/0336f96af020afb62f6de616b19196d0cde711c5/internal/util/endpoints.go#L108), create a new HTTP client, stored as a global variable, and explicit set a default timeout, perhaps ~10 seconds.

---

* **Experience**: Does the project provide a good experience to users?
* **Internals**: Is the implementation efficient, effective, and maintainable?

SaketramDurbha:
This PR fixes #16.

It adds a timeout of 10 seconds to HTTP requests made to Cloud Run services.

An option for users to configure the duration of this timeout (#17) will come in the future.

### DIFF
--- a/internal/util/endpoints.go
+++ b/internal/util/endpoints.go
@@ -22,12 +22,19 @@ import (
 	"net/http"
 	"strconv"
 	"strings"
+	"time"
 )
 
 // test holds the openapi3.Operation and HTTP method associated with a single
 type test struct {
 	operation  *openapi3.Operation
 	httpMethod string
+}
+
+// httpClient is the client that used for making HTTP requests to Cloud Run services. It is the same as the
+// http.DefaultClient, except for the fact that it has a timeout for requests made.
+var httpClient = &http.Client{
+	Timeout: 10 * time.Second,
 }
 
 // ValidateEndpoints tests all paths (represented by openapi3.Paths) with all HTTP methods and given response bodies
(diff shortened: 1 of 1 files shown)


kurtisvg (approved):

internal/util/endpoints.go
```diff
@@ -30,6 +31,12 @@ type test struct {
 	httpMethod string
 }
 
+// httpClient is the client that used for making HTTP requests to Cloud Run services. It is the same as the
+// http.DefaultClient, except for the fact that it has a timeout for requests made.
+var httpClient = &http.Client{
```
tritone: This is probably fine here, but I'd say it's more typical to do timeouts on a per-request basis using a context. Usually you would be making multiple types of requests with a single http.Client and would want more granular control on individual timeouts. To do this:
* Create a context with a timeout using `WithTimeout`
* Use https://golang.org/pkg/net/http/#Request.WithContext to attach this to the request passed to your client.

tritone (approved):
One suggestion, up to you if you implement or not I would say.

internal/util/endpoints.go
grayside: Being of a defensive mindset I prefer to have a sane default value for this, but no objection to the other approach. One consideration is to keep the existing implementation, but use WithTimeout in the HTTP requests to apply overrides as needed.

An interesting follow-up might be to read a per operation timeout from the configuration along the lines of this custom OpenAPI Extension: https://cloud.google.com/endpoints/docs/openapi/openapi-extensions#deadline

SaketramDurbha: Resolved, thank you both!

With regards to configurable timeouts for each test, we haven't added the option yet for users to specify their own OpenAPI file yet (we're currently only doing a default GET request to the root endpoint and expecting a 200 status code), but we'll keep Google's OpenAPI Extension in mind while implementing it.

I've added a TODO in the code for making the timeouts user-configurable per-test.

kurtisvg (approved):

tritone (approved):

PR #27, opened 2020-08-18
add support for Cloud Build config (resolve #8)

### LINKED ISSUE #8: Add support for "cloudbuild.yaml" as a build mechanism or log a warning
If `cloudbuild.yaml` is defined for a given sample that means build, deployment, and possibly CI and teardown are all defined there.

One of the following should be done:
* Add support for Cloud Build as highest build priority short of a custom override
* Have sts throw a warning if it sees a `cloudbuild.yaml` as it may be running a build/deploy different from what's intended by the project.

---

* **Scope**: Does the tool have the breadth of features and flexibility to support reasonable testing needs?

SaketramDurbha:
This issue resolves Issue #8.

Currently set up with making a mandatory substitution `_SST_RUN_REGION` for passing the Cloud Run region to the Cloud Build config file, but this is open to being changed.


internal/lifecycle/lifecycle.go
```diff
@@ -111,3 +134,46 @@ func buildDefaultJavaLifecycle(serviceName, gcrURL string) Lifecycle {
 
 	return l
 }
+
+// replaceServiceName takes a terminal command's arguments as input and replaces the Cloud Run service name, if any.
+// It detects whether the command is a gcloud run command and replaces the last argument that isn't a flag
+// with the input service name.
+func replaceServiceName(name string, args []string, serviceName string) error {
```
SaketramDurbha, line 151: This is a placeholder — @danielylee01 is working on an updated version of this.

README.md
```diff
@@ -63,5 +83,7 @@ In addition to setting a default Cloud Run region, make sure to deploy to the fu
 can achieve this by setting the `run/platform` gcloud property to `managed` or passing in the `--platform=managed` flag
 to your `gcloud run` commands.
 
-If comment code tags aren't added to your README, the program will fall back to reasonable defaults to build and deploy
-your sample to Cloud Run based on whether your sample is java-based and doesn't have a Dockerfile or isn't.
+### Reasonable defaults
```
kurtisvg:
```suggestion
### Defaults
```

I'd suggest making this section a list. `Be default, SST will attempt to get deploy instructions from the following locations:`

internal/lifecycle/cloud_build_config.go
```diff
@@ -0,0 +1,111 @@
...
+			arg := config["steps"].([]interface{})[stepIndex].(map[interface{}]interface{})["args"].([]interface{})[argIndex].(string)
+			arg = gcrURLRegexp.ReplaceAllString(arg, gcrURL)
+
+			args = append(args, arg)
+		}
+
+		prog := config["steps"].([]interface{})[stepIndex].(map[interface{}]interface{})["name"].(string)
+		err := replaceServiceName(prog, args, serviceName)
+		if err != nil {
+			return nil, fmt.Errorf("[lifecycle.parseCloudBuildConfig] replacing Cloud Run service name in Cloud Build config step args: %w", err)
```
kurtisvg: It's weird to keep reviewing these `[func_name]` after your other PR to change them. You should probably start doing it the new way for new PRs, so that PR only covers existing code and doesn't constantly need to be re-reviewed.

internal/cmd/cmd.go
```diff
@@ -19,11 +19,22 @@ import (
 	"github.com/GoogleCloudPlatform/serverless-sample-tester/internal/sample"
 	"github.com/GoogleCloudPlatform/serverless-sample-tester/internal/util"
 	"github.com/spf13/cobra"
+	"github.com/spf13/viper"
 	"log"
 	"os/exec"
 	"path/filepath"
 )
 
+// InitConfig initializes binds command line flags and environment variables to viper configuration keys. This should
```
SaketramDurbha: This is also a placeholder for @danielylee01 's work, to be built off of this PR: https://github.com/GoogleCloudPlatform/serverless-sample-tester/pull/22

internal/util/tempfiles.go
```diff
@@ -0,0 +1,47 @@
...
+	tempFile, err := ioutil.TempFile("", "example")
+	if err != nil {
+		return tempFile, fmt.Errorf("[util.CreateTempFile] creating temp file: %w\n", err)
+	}
+
+	tempFiles = append(tempFiles, tempFile)
+	return tempFile, nil
+}
+
+func RemoveTempFiles() error {
```
kurtisvg: You have an error return type here, but you only ever return `nil`. Is this intentional?

internal/util/tempfiles.go
```diff
@@ -0,0 +1,47 @@
...
+	"io/ioutil"
+	"log"
+	"os"
+)
+
+// tempFiles is a slice of pointers to temporary files created throughout the course of the program. Users should call
+// RemoveTempFiles at the end of the program.
+var tempFiles []*os.File
+
+func CreateTempFile() (*os.File, error) {
```
kurtisvg: So this uses global state, and as such won't be threadsafe. 

A common pattern for this is to return a `closer` function that can cleanup after the resource is no longer needed. For example:
```go
tmp, cleanup, err := CreateTempFile()
if err != nil {

}
defer cleanup()
```
This is nice because you can defer it, making sure it gets cleaned up even if something goes wrong. It also helps scope the resource to exactly when it's needed in the function, instead of waiting for cleanup at the end. (and avoids the dangers of having a global state, like not being threadsafe etc). 

However, once you remove the `tempFiles` this is pretty much just the following:
```go
tempFile, err := ioutil.TempFile("", "example")
if err != nil {
	return tempFile, fmt.Errorf("[util.CreateTempFile] creating temp file: %w\n", err)
}
defer os.RemoveAll(tempFile)
```

So in that case a utility function may no longer be helpful.

internal/lifecycle/lifecycle.go
```diff
@@ -46,7 +51,24 @@ func (l Lifecycle) Execute(commandsDir string) error {
...
 // (has a pom.xml) that doesn't have a Dockerfile or isn't.
-func NewLifecycle(sampleDir, serviceName, gcrURL string) (Lifecycle, error) {
+func NewLifecycle(sampleDir, serviceName, gcrURL, runRegion string, cloudBuildConfSubs map[string]string) (Lifecycle, error) {
+	// First try Cloud Build Config file
+	cloudBuildConfigPath := fmt.Sprintf("%s/cloudbuild.yaml", sampleDir)
+
+	_, err := os.Stat(cloudBuildConfigPath)
+	cloudBuildConfigE := err == nil
+
+	if cloudBuildConfigE {
```
kurtisvg: Any reason not to just inline this? 

```suggestion
	if err == nil {
```

kurtisvg (changes requested):

internal/util/tempfiles.go
SaketramDurbha: Resolved since we've decided to remove these functions. Thank you!

internal/lifecycle/lifecycle.go
SaketramDurbha: Resolved, thank you.

README.md
SaketramDurbha: Addressed this in a recent commit; was this what you were looking for?

internal/lifecycle/cloud_build_config.go
SaketramDurbha: Resolved, thank you.

internal/util/tempfiles.go
SaketramDurbha: Resolved, thank you.

kurtisvg (approved):

PR #28, opened 2020-08-18
fix default build mechanism priority for jib (address #7)

### LINKED ISSUE #7: Fix default build mechanism priority for jib
Default build priority doesn't cover all the supported build methods, and incorrectly prioritizes Dockerfiles over Jib for Java.

## Current Practice

- README command
- Dockerfile
- Jib

## Preferred

- README command
- Jib
- Dockerfile

Notes on the preferred ordering:

- Jib is the primary recommendation for building Java containers in Cloud Run. Jib may be configured in a project to use a Dockerfile to create a base image.

This raises a secondary problem: what if it's a Java project that wants to build from a Dockerfile, not use jib?

* **Correctness**: Is the implemented business logic correct?
* **Scope**: Does the tool have the breadth of features and flexibility to support reasonable testing needs?

SaketramDurbha:
This PR addresses issue #7 in part.

For all Java-based samples (`pom.xml` in the sample's root directory), we're going to use Jib, regardless of whether or not the sample also has a `Dockerfile` included in its root directory.

If we want, we can also include a `—no-jib` flag here that would force the program to use `gcloud builds submit` for all samples.

### DIFF
--- a/README.md
+++ b/README.md
@@ -91,4 +91,9 @@ then `$CLOUD_RUN_SERVICE_NAME` should be set to `run-mysql`.
 
 ### Defaults
 If comment code tags aren't added to your README, the program will fall back to reasonable defaults to build and deploy
-your sample to Cloud Run based on whether your sample is Java-based (has a `pom.xml` in its root directory) or not.
+your sample to Cloud Run based on whether your sample is Java-based (has a pom.xml) or not.
+
+If your sample is Java-based, your sample will be built and pushed to the Container Registry using `mvn compile com.google.cloud.tools:jib-maven-plugin:2.0.0:build -Dimage=tag`.
+Otherwise, your sample will be built and pushed to the Container Registry using `gcloud builds submit --tag=[tag]`. 
+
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -63,5 +63,12 @@ In addition to setting a default Cloud Run region, make sure to deploy to the fu
 can achieve this by setting the `run/platform` gcloud property to `managed` or passing in the `--platform=managed` flag
 to your `gcloud run` commands.
 
+### Defaults
+
 If comment code tags aren't added to your README, the program will fall back to reasonable defaults to build and deploy
-your sample to Cloud Run based on whether your sample is java-based and doesn't have a Dockerfile or isn't.
+your sample to Cloud Run based on whether your sample is Java-based (has a pom.xml) or not.
```
kurtisvg, line 94: See my other comments about having this be a list

kurtisvg (approved):

README.md
SaketramDurbha: Resolved, thank you.

kurtisvg (approved):

Issue #19, opened 2020-08-10
Revisit directory structure and package scoping


grayside:
Package naming and scope seems like it's been slightly outgrown by the way the code has been implemented. As a result, it's difficult to know where to look to see code for the different aspects of the tool.

A few examples of this are gathered here:

* Testing is a main feature of this tool, but the testing code is implemented inside `util` package.
* `util/config.go` seems to be for OpenAPI file parsing. A more specific name would be helpful.
* Were I to guess what the `lifecycle` package is about, I'd expect to find organization of all build, deploy, test, and teardown steps. However, it's only build.
* The `gcloud` package is in fact a subset of Cloud Run specific operations, the use of gcloud is an implementation detail.
* Because of the lack of clarity in some of the other package names, I find it difficult to recall what might be found in `sample`.

Happy to discuss possible code organization improvements.

---

This issue affects two categories in my review of the repository:
* **Adoption**: Can this project be found and does it provide enough context on when and how to use it?  Does it inspire confidence as a well-maintained project?
* **Internals**: Is the implementation efficient, effective, and maintainable?

PR #30, opened 2020-08-21
Reworked directory tree and main command flow

danielylee01:
In response to #19, this change addresses some clutter that was forming in the directory. The duplicate cmd directories were merged, and the main logical flow was moved to `root.go`. 

In particular, it lays groundwork for adding flags to the tool. As described [here](https://docs.google.com/document/d/1VuC2bQK2-zUzulMagwq3enQ8QURfRGprosPoAL-Ep0A/edit?usp=sharing), flags will be immediately useful for allowing the user to manually override the region. The tool now more closely mirrors the documentation in [spf13/cobra documentation](https://github.com/spf13/cobra), increasing legibility for future work.


internal/cmd/root.go
```diff
@@ -0,0 +1,108 @@
+// Copyright 2020 Google LLC
```
kurtisvg: Why is this file in `internal`? This file cmd related, so I'd expect it to just be in `cmd`

internal/cmd/root.go
```diff
@@ -0,0 +1,108 @@
...
+	return rootCmd.Execute()
+}
+
+// init initializes the tool.
+func init() {
+	cobra.OnInitialize(initConfig)
+}
+
+func initConfig() {
+	// Support for tool config file. Waiting until config functionality is added to
```
kurtisvg: Probably just remove this function altogether and add this bit in when the other PR gets merged

kurtisvg (changes requested):

cmd/root.go
```diff
@@ -0,0 +1,94 @@
...
+			log.Println("Loading test endpoints")
+			swagger := util.LoadTestEndpoints()
+
+			log.Println("Building and deploying sample to Cloud Run")
+			err = s.BuildDeployLifecycle.Execute(s.Dir)
+			defer s.Service.Delete(s.Dir)
+			defer s.DeleteCloudContainerImage()
+			if err != nil {
+				return fmt.Errorf("[cmd.Root] building and deploying sample to Cloud Run: %w", err)
+			}
```
kurtisvg: Generally you don't cleanup unless you are sure these things exist. Are these cleanups safe if there is nothing to clean up, or should we move this below err?

cmd/root.go
```diff
@@ -0,0 +1,94 @@
...
+
+			log.Println("Getting identity token for gcloud auhtorized account")
+			var identToken string
+
+			a := append(util.GcloudCommonFlags, "auth", "print-identity-token")
+			identToken, err = util.ExecCommand(exec.Command("gcloud", a...), s.Dir)
+
```
kurtisvg: nit: newline probably isn't neccesary

cmd/root.go
```diff
@@ -0,0 +1,94 @@
...
+			err = s.BuildDeployLifecycle.Execute(s.Dir)
+			defer s.Service.Delete(s.Dir)
+			defer s.DeleteCloudContainerImage()
+			if err != nil {
+				return fmt.Errorf("[cmd.Root] building and deploying sample to Cloud Run: %w", err)
+			}
+
+			log.Println("Getting identity token for gcloud auhtorized account")
+			var identToken string
+
```
kurtisvg: nit: newline probably isn't necessary

kurtisvg (approved):
3 nits, otherwise LGTM

cmd/root.go
danielylee01: These cleanups are indeed safe even if there's nothing to clean up. Additionally, since this runs after the Lifecycle occurs, we want to be sure that we're still cleaning up the resources in case there's an error only in one of the later stages and not, say, the build stage.

PR #31, opened 2020-08-22
follow go style of table-driven tests more closesly

SaketramDurbha:
In response to: https://github.com/GoogleCloudPlatform/serverless-sample-tester/pull/25#discussion_r475034303

Use `T.Run` for creating subtests for each test case.

### DIFF
--- a/internal/lifecycle/readme_test.go
+++ b/internal/lifecycle/readme_test.go
@@ -3,6 +3,7 @@ package lifecycle
 import (
 	"bufio"
 	"errors"
+	"fmt"
 	"os"
 	"os/exec"
 	"reflect"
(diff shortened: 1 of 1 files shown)


internal/lifecycle/readme_test.go
```diff
@@ -193,40 +194,42 @@ var toCommandsTests = []toCommandsTest{
 
 func TestToCommands(t *testing.T) {
 	for i, tc := range toCommandsTests {
-		if len(tc.codeBlock) == 0 {
-			continue
-		}
+		t.Run(fmt.Sprintf("#%d", i), func(t *testing.T) {
```
tritone: Instead of just giving a number here, your first arg to `T.Run` should be a test case description. Add a name field to `toCommandsTest` and fill that with a brief description, replacing the comments that you have added to each case. That way when there's a failure it's clear from the output which test case failed.

internal/lifecycle/readme_test.go
```diff
@@ -250,22 +253,23 @@ var parseREADMETests = []parseREADMETest{
 
 func TestParseREADME(t *testing.T) {
 	for i, tc := range parseREADMETests {
-		if tc.inFileName == "" {
-			continue
-		}
+		t.Run(fmt.Sprintf("#%d", i), func(t *testing.T) {
```
tritone: Same idea as above here.

SaketramDurbha: Resolved, thank you.

internal/lifecycle/readme_test.go
SaketramDurbha: Resolved, thank you. Does the updated version have the description format you were looking for?

SaketramDurbha:
Also uncommenting a test case that was originally commented out, addressed by #29.

internal/lifecycle/readme_test.go
```diff
@@ -38,25 +39,25 @@ const uniqueServiceName = "unique_service_name"
...
 var toCommandsTests = []toCommandsTest{
-	// single one-line command
 	{
+		description: "single one-line command test",
 		codeBlock: codeBlock{
 			"echo hello world",
 		},
 		cmds: []*exec.Cmd{
-			exec.Command("echo", "hello", "world"),
+			exec.Command("echo", "hello", "wosrld"),
```
tritone: typo? did you run this locally?

internal/lifecycle/readme_test.go
```diff
@@ -38,25 +39,25 @@ const uniqueServiceName = "unique_service_name"
...
+	codeBlock   codeBlock         // input code block
+	cmds        []*exec.Cmd       // expected result of codeBlock.toCommands
+	err         string            // expected string contained in return error of codeBlock.toCommands
+	env         map[string]string // map of environment variables to values for this test
 }
 
 var toCommandsTests = []toCommandsTest{
-	// single one-line command
 	{
+		description: "single one-line command test",
```
tritone: You can leave "test" off the end of all of these.

tritone (approved):
Generally looks good, two small things. Run `go test -v` to see the full output and confirm that it's what you're expecting.

tritone (approved):

